PR #12, opened 2016-09-26
(wip) add method overloading to isEnabled

sveisvei:


src/client.ts
```diff
@@ -58,7 +61,7 @@ export default class UnleashClient {
                     this.errorHandler(new Error(`Missing strategy ${strategySelector.name}`));
                     return false;
                 }
-                return strategy.isEnabled(strategySelector.parameters, context);
+                return strategy.isEnabled(strategySelector.parameters, context, defaultValue);
```
ivarconr: I don't think the default value has anything to do with the strategy. The idea of the default value is to use that if someone calls `isEnabled("name", {}, fale)` for an unknown feature toggle (because the client has not gotten anything about it from the unleash server. This is handy in case of kill-switches etc.

PR #13, opened 2016-09-27
add fallbackValue stupport to isEnabled

sveisvei:
cc @ivarconr

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -37,8 +37,12 @@ export default class UnleashClient {
         return match;
     }
 
-    isEnabled (name: string, context: any) : boolean {
+    isEnabled (name: string, context: any, fallbackValue?: boolean) : boolean {
         const feature: FeatureInterface = this.repository.getToggle(name);
+
+        if (!feature && typeof fallbackValue === 'boolean') {
+            return fallbackValue;
+        }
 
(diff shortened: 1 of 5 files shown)


ivarconr:
Oh, this is actually a different solution than I was thinking about. I was thinking of two cases: 

1. unleash is not initialized yet
2. the feature toggle is not defined yet in unleash-server

This could be solved with the following check inside `client.js`:

```js
if (!feature) {
  return defaultValue;
}
```

sveisvei:
this is to solve the point 1).

Can add support for point 2) as well

ivarconr:
not sure if my solution is better though, but it's kinda the way it was solved in the javaclient: 

https://github.com/finn-no/unleash-client-java/blob/master/src/main/java/no/finn/unleash/DefaultUnleash.java#L42

```java
        if (featureToggle == null) {
            return defaultSetting;
        } else if(!featureToggle.isEnabled()) {
            return false;
        }
```

ivarconr:
excellent point!

sveisvei:
Added support for 2 (missing test atm), is this what you think is the complete solution?

ivarconr:
yes, this nails it! :+1: :zap:

PR #14, opened 2016-09-27
- Enabled the strictNullChecks flag when compiling

runeh:
- Added explicit null checking a few places where neccessary
- Removed some null checks in places where the compiler should tells us about nullability


sveisvei:
💯  Thanx

PR #15, opened 2016-10-11
Fix install

SimenB:
Without this change, it fails on install with:

```
> unleash-client@1.0.0-alpha.1 build /Users/simbekkh/repos/express-base/node_modules/unleash-client
> tsc src/*.ts -d --sourceMap --outdir ./lib --lib ES6,ES5

sh: tsc: command not found
```

If I install typescript, it still fails because the source code is not included in the package.

```
> unleash-client@1.0.0-alpha.1 build /Users/simbekkh/repos/express-base/node_modules/unleash-client
> tsc src/*.ts -d --sourceMap --outdir ./lib --lib ES6,ES5

error TS6053: File 'src/*.ts' not found.
```

Also allow this to be run on Node@7 without warnings (release imminent)

/cc @sveisvei @ivarconr

### DIFF
--- a/examples/custom_strategy.js
+++ b/examples/custom_strategy.js
@@ -13,7 +13,7 @@ class ActiveForUserWithEmailStrategy extends Strategy {
 
 const client = initialize({
     url: 'http://unleash.herokuapp.com/features',
-    refreshIntervall: 10000,
+    refreshInterval: 10000,
     strategies: [new ActiveForUserWithEmailStrategy()],
 });
 
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "request": "^2.74.0"
   },
   "engines": {
-    "node": "6"
+    "node": ">=6"
   },
   "files": [
     "lib",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "eslint .",
     "build": "tsc src/*.ts -d --strictNullChecks --sourceMap --outdir ./lib --lib ES6,ES5",
-    "postinstall": "npm run build",
+    "prepublish": "not-in-install && npm run build || in-install",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "request": "^2.74.0"
   },
   "engines": {
-    "node": "6"
+    "node": ">=6"
```
SimenB: You test on node 4, though, why lock it to 6?

src/repository.ts
```diff
@@ -11,7 +11,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
     private storage: Storage;
     private etag: string;
 
-    constructor (backupPath: string, url: string, refreshIntervall?: number, StorageImpl = Storage) {
+    constructor (backupPath: string, url: string, refreshInterval?: number, StorageImpl = Storage) {
```
SimenB: Breaking change, I suppose...

examples/custom_strategy.js
```diff
@@ -13,7 +13,7 @@ class ActiveForUserWithEmailStrategy extends Strategy {
 
 const client = initialize({
     url: 'http://unleash.herokuapp.com/features',
-    refreshIntervall: 10000,
+    refreshInterval: 10000,
```
sveisvei: This client is not widely used as far as I know, so not worth it imo.

sveisvei:
Add a console.log, and this is fine to merge and release another alpha.

SimenB:
Just log, or map it over from the old name as well?

src/repository.ts
SimenB: 👍

sveisvei:
On second thought, this is a breaking change, lets throw instead as its startup time.

SimenB:
Kk!

PR #17, opened 2016-10-14
Remove silly url check

SimenB:
k8s for instance is just `unleash-server`


SimenB:
This is blocking for using Unleash on fiaas

PR #20, opened 2016-10-26
Metrics

sveisvei:
* Breaking: requires `appName` to be provided
* register client with strategynames at /client/register on startup
* report 60 seconds buckets of metrics to /client/metrics on unleash server
* exposing metric events

cc @ivarconr @SimenB

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -55,6 +55,10 @@ export default class UnleashClient {
             return false;
         }
 
+        if (feature.strategies.length === 0) {
+            return feature.enabled;
+        }
+
         return feature.strategies.length > 0 && feature.strategies
             .some((strategySelector) : boolean => {
                 const strategy: Strategy = this.getStrategy(strategySelector.name);
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -69,7 +69,7 @@ export default class Metrics extends EventEmitter {
 
     stop () {
         clearInterval(this.timer);
-        this.timer = null;
+        delete this.timer;
         this.disabled = true;
     }
 
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -50,28 +50,38 @@ export default class Metrics extends EventEmitter {
         this.started = new Date();
         this.resetBucket();
 
-        if (this.metricsInterval > 0) {
+        if (typeof this.metricsInterval === 'number' && this.metricsInterval > 0) {
             this.startTimer();
             this.registerInstance();
         }
     }
 
     private startTimer () {
+        if (this.disabled) {
+            return false;
+        }
         this.timer = setTimeout(() => {
             this.sendMetrics();
         }, this.metricsInterval);
         this.timer.unref();
-
+        return true;
     }
 
-    registerInstance () {
+    stop () {
+        clearInterval(this.timer);
+        this.timer = null;
+        this.disabled = true;
+    }
+
+    registerInstance () : boolean {
         if (this.disabled) {
-            return;
+            return false;
         }
         const url = resolve(this.url, '/client/register');
+        const payload = this.getClientData();
         post({
             url,
-            json: this.getClientData(),
+            json: payload,
         }, (err, res: ClientResponse, body) => {
             if (err) {
                 this.emit('error', err);
@@ -82,21 +92,25 @@ export default class Metrics extends EventEmitter {
                 this.emit('warn', `${url} returning ${res.statusCode}`);
                 return;
             }
+            this.emit('registered', payload);
         });
+        return true;
     }
 
-    sendMetrics () {
+    sendMetrics () : boolean {
         if (this.disabled) {
-            return;
+            return false;
         }
         if (this.bucketIsEmpty()) {
             this.resetBucket();
-            return this.startTimer();
+            this.startTimer();
+            return true;
         }
         const url = resolve(this.url, '/client/metrics');
+        const payload = this.getPayload();
         post({
              url,
-             json: this.getPayload(),
+             json: payload,
         }, (err, res: ClientResponse, body) => {
             this.startTimer();
             if (err) {
@@ -106,7 +120,7 @@ export default class Metrics extends EventEmitter {
 
             if (res.statusCode === 404) {
                 this.emit('warn', `${url} returning 404, stopping metrics`);
-                clearTimeout(this.timer);
+                this.stop();
                 return;
             }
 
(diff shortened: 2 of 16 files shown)


src/metrics.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    count (name: string, enabled: boolean) : void {
+        if (!this.store[name]) {
+            this.store[name] = {
+                yes: [],
+                no: [],
+            };
+        }
+        this.store[name][enabled ? 'yes' : 'no'].push({
+            time: new Date(),
```
SimenB: `Date.now`?

src/metrics.ts
```diff
@@ -0,0 +1,126 @@
...
+
+
+    count (name: string, enabled: boolean) : void {
+        if (!this.bucket.toggles[name]) {
+            this.bucket.toggles[name] = {
+                yes: 0,
+                no: 0,
+            };
+        }
+        this.bucket.toggles[name][enabled ? 'yes' : 'no'] += 1;
```
SimenB: `++`?

src/metrics.ts
```diff
@@ -0,0 +1,126 @@
...
+
+            if (res.statusCode !== 200) {
+                this.emit('warn', `${url} returning ${res.statusCode}`);
+                return;
+            }
+        });
+    }
+
+
+    count (name: string, enabled: boolean) : void {
```
SimenB: I'd love to see this emit the fact it counted 😄

src/unleash.ts
```diff
@@ -34,18 +42,51 @@ export class Unleash extends EventEmitter {
...
+        this.repository.on('error', (err) => {
+            err.message = `Unleash Repository error: ${err.message}`;
+            this.emit('error', err);
+        });
+
+        this.repository.on('warn', (msg) => {
+            this.emit('warn', msg);
+        });
+
+        this.metrics = new Metrics({
```
ivarconr: we might want a flag to disable metrics in case it prove to be some perf issues?

sveisvei: inc flag

src/metrics.ts
sveisvei: Whats the usecase? :)

SimenB: https://github.com/Unleash/unleash-client-node/pull/20#pullrequestreview-6004982

> so we can add prometheus on the consumer side

src/client.ts
```diff
@@ -55,6 +55,10 @@ export default class UnleashClient {
             return false;
         }
 
+        if (feature.strategies.length === 0) {
```
sveisvei: This is somewhat unreleasted @ivarconr, but should be default behaviour?

SimenB:
Why does the client report to the server, instead of the server asking for metrics on demand? Sorta like prometheus does it, you just say "hey, I've got metrics!" and the server then polls it when it wants to.

src/metrics.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (this.disabled) {
+            return false;
+        }
+        if (!this.bucket.toggles[name]) {
+            this.bucket.toggles[name] = {
+                yes: 0,
+                no: 0,
+            };
+        }
+        this.bucket.toggles[name][enabled ? 'yes' : 'no'] ++;
```
SimenB: space before `++`? Is that our eslint rule?

src/request.ts
```diff
@@ -0,0 +1,36 @@
+import * as request from 'request';
```
SimenB: Why not a promise based one? As we're close to async await, that's cleaner, IMO

src/unleash.ts
```diff
@@ -9,8 +10,12 @@ import { EventEmitter } from 'events';
 const BACKUP_PATH: string = tmpdir();
 
 export interface UnleashConfig {
+    appName?: string,
```
SimenB: should it be `?` if you throw if it's missing?

src/unleash.ts
```diff
@@ -34,32 +44,67 @@ export class Unleash extends EventEmitter {
...
+        this.metrics = new Metrics({
+            disableMetrics,
+            appName,
+            instanceId,
+            strategies: strategies.map((strategy: Strategy) => strategy.name),
+            metricsInterval,
+            url
+        });
+
+        this.metrics.on('error', (err) => {
```
SimenB: still missing an `this.metrics.on(''count')` so we can also chuck this in the consumers of the client. Seems like only I want that though

test/metrics.network.test.js
```diff
@@ -0,0 +1,40 @@
+import test from 'ava';
+import Metrics from '../lib/metrics';
+import nock from 'nock';
+
+nock.disableNetConnect();
```
SimenB: should be in a `before`, and removed in an `after`, or your tests have side effects

test/metrics.test.js
```diff
@@ -0,0 +1,197 @@
+import test from 'ava';
+import Metrics from '../lib/metrics';
+import nock from 'nock';
+
+let counter = 1;
+const getUrl = () => `http://test${counter++}.app/`;
```
SimenB: side efects? not very ava-ish

test/unleash.test.js
```diff
@@ -6,6 +6,9 @@ import mkdirp from 'mkdirp';
 
 import { Unleash } from '../lib/unleash';
 
+let counter = 1;
+const getUrl = () => `http://test2${counter++}.app/`;
```
SimenB: again, not a fan of this side effect

SimenB:
Some nits, but overall LGTM!

test/metrics.network.test.js
sveisvei: Indeed

src/metrics.ts
sveisvei: Sure!

src/unleash.ts
sveisvei: Will add

test/metrics.test.js
sveisvei: I dont want nock to interfere cross tests, could do random instead()?

sveisvei:
Merged this. 

Im not 100% sure on the event names from the metrics @SimenB , if any have better suggestions, let me know.

src/metrics.ts
sveisvei: This is typescript. I have not hooked up tslint atm.

src/request.ts
sveisvei: Did test, but code did not look better with promises on the request part.

src/unleash.ts
sveisvei: Good catch!

ivarconr:
@SimenB we can not do what prometheus do for multiple reasons:

- we can not enforce all users to expose a port for us (we might move in direction of unleash hosted at some point)
- not all clients are using HTTP (also thrift services are using unleash)
- It would be a total redesign. I am more in favor of moving in direction of websockets or similar to mininmize the load.

PR #21, opened 2016-11-06
remove errorHandler, use error event instead

sveisvei:
* having a custom errorHandler was not intended

cc @ivarconr


ivarconr:
looks good!

PR #29, opened 2016-12-06
emit warning once on missing strategy

### LINKED ISSUE #28: Throttle warnings on missing strategy
We should not warn on every missing strategy? But we need to surface somehow - so maybe a throttle solution might be best?

sveisvei:
Closes #28

cc @SimenB @ivarconr


ivarconr:
:+1:

SimenB:
Perf! Could consider emitting as info after initial warn, but this makes a lot of sense :smile:

Issue #35, opened 2017-04-24
Unhandled error when initializing client for unknown user


jrbarron:
This was discovered while running autotests for our app so is probably not critical to production.

If the Unleash client is initialized without specifying an `instanceId` then the client will attempt to generate one using `os.userInfo` and `os.hostname`. This can result in an unhandled error in some cases because `os.userInfo` will throw an error if the user running the process does not have an entry in the `/etc/passwd` file on *nix systems. Here is a stack trace from our app:

```
/usr/src/app/node_modules/unleash-client/lib/unleash.js:44
            var info = os_2.userInfo();
                            ^

Error: ENOENT: no such file or directory, uv_os_get_passwd
    at Error (native)
    at new Unleash (/usr/src/app/node_modules/unleash-client/lib/unleash.js:44:29)
    at initialize (/usr/src/app/node_modules/unleash-client/lib/index.js:8:16)
```

See `GetUserInfo` in Node's os module:

https://github.com/nodejs/node/blob/master/src/node_os.cc#L354

Since there is already a fallback for the case where `info.username` is empty, it probably makes sense to put a try/catch around the call to userInfo() and ensure the fallback gets used if an error is thrown.

In the meantime, the workaround is simply to pass in a `instanceId` when initializing Unleash.

ivarconr:
Ai, thanks. Fortunately it should be easy to provide a fix for this one!

ivarconr:
fixed in 2.1.2

PR #36, opened 2017-04-27
Unleash initalize should not fail when os.userInfo() throws #35

ivarconr:


ivarconr:
@jrbarron I believe this solves the problem. 

I also want to add an test for this, but need to figure out how. Any suggestion on how to make os.userInfo() fail in a test @sveisvei?

jrbarron:
@ivarconr Yup, your fix is the same that I had applied locally so I can confirm that it fixes the issue for us :)

PR #41, opened 2017-06-15
Run tests on stable node, and use yarn for commands

SimenB:


.travis.yml
```diff
@@ -1,21 +1,18 @@
 sudo: false
 language: node_js
 node_js:
-- "6"
-- "7"
+- 6
+- stable
 script:
-- npm install
```
SimenB: ran automatically

SimenB:
Fails because of nock, which doesn't work on node 8.

SimenB:
node-nock/nock#922

SimenB:
@sveisvei there is an actual failing test on node 8, could you take a look?

PR #44, opened 2017-06-19
Include data in metrics

SimenB:


ivarconr:
Nice catch!

src/metrics.ts
```diff
@@ -133,7 +133,7 @@ export default class Metrics extends EventEmitter {
                 }
 
                 if (!(res.statusCode && res.statusCode >= 200 && res.statusCode < 300)) {
-                    this.emit('warn', `${url} returning ${res.statusCode}`);
+                    this.emit('warn', `${url} returning ${res.statusCode}`, body);
```
ivarconr: :+1:

src/metrics.ts
```diff
@@ -1,5 +1,5 @@
 import { EventEmitter } from 'events';
-import { ClientResponse } from 'http';
+import { IncomingMessage } from 'http';
```
SimenB: It was deprecated, so just changed to what it told me to

PR #47, opened 2017-06-28
Upgrade ava

SimenB:


package.json
```diff
@@ -52,13 +52,16 @@
...
+    "statements": 98,
+    "functions": 97,
+    "branches": 93,
     "include": [
       "lib/**.js"
     ],
     "all": true
+  },
+  "ava": {
+    "failWithoutAssertions": false
```
sveisvei: I kinda like this being on though... you think its annoying?

SimenB: There is a test without assertions now

sveisvei: then use `t.plan(0)` :)

SimenB:
One failing test, unsure what's up with it?

sveisvei:
@SimenB force pushing here fyi.

sveisvei:
@SimenB merge if you are OK with the `process.env.NODE_ENV=test` solution

SimenB:
Hah, this was https://github.com/avajs/ava/issues/1484?

package.json
```diff
@@ -9,7 +9,7 @@
     "lint:fix": "eslint . --fix  && prettier --write '{src,examples}/**/*.ts' --single-quote --tab-width 4 --print-width 100 --trailing-comma all --parser typescript",
     "build": "tsc src/*.ts src/**/*.ts -d --strictNullChecks --sourceMap --outdir ./lib --lib ES6,ES5",
     "prepublish": "not-in-install && npm run build || in-install",
-    "test": "ava test",
+    "test": "NODE_ENV=test ava test",
```
SimenB: Won't work on windows without doing `cross-env` or something. maybe care

sveisvei: I dont - but would if I knew of anyone running these tests on a windows env.

PR #58, opened 2017-11-17
adding IP range support to remoteAddress

leftieFriele:
Adding a gazilion IPs manually is somewhat of a pain, therefor adding support for IPv4 ranges would really make things easier. 
This PR uses the ip module to validate the ranges. I have not updated the docs on this as I wasn't sure if that should go into the PR.

### DIFF
--- a/src/strategy/remote-addresss-strategy.ts
+++ b/src/strategy/remote-addresss-strategy.ts
@@ -12,14 +12,14 @@ export class RemoteAddressStrategy extends Strategy {
             return false;
         }
         for (const range of parameters.IPs.split(/\s*,\s*/)) {
-            if (!ip.isV6Format(range)) {
+            if (range === context.remoteAddress) {
+                return true;
+            } else if (!ip.isV6Format(range)) {
                 if (ip.cidrSubnet(range).contains(context.remoteAddress)) {
                     return true;
                 }
             }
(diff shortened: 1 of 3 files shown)


src/strategy/remote-addresss-strategy.ts
```diff
@@ -10,7 +11,16 @@ export class RemoteAddressStrategy extends Strategy {
...
         }
-        const ipList = parameters.IPs.split(/\s*,\s*/);
-        return ipList.includes(context.remoteAddress);
+        for (const range of parameters.IPs.split(/\s*,\s*/)) {
+            if (!ip.isV6Format(range)) {
+                if (ip.cidrSubnet(range).contains(context.remoteAddress)) {
+                    return true;
+                }
+            }
+            if (range === context.remoteAddress) {
```
ivarconr: Minor: Maybe this should the the if `if` statement. Feels like this could be faster than the range check?

ivarconr:
Awesome :fire: . 

Range-support has been on my wishlist for a while. We should port it to the other client implementations as well. 

I can release this as `2.3.1` pretty fast, as soon as we settle this PR.

ivarconr:
released as 2.3.1 (https://www.npmjs.com/package/unleash-client)

PR #67, opened 2018-01-16
RemoteAddressStrategy should ignore invalid IPs.

### LINKED ISSUE #38: Throw/warn if multiple versions loaded

ivarconr:
Today the strategy ends up throwing, making the
whole isEnable call throw if the user has specified
an invalid IP.

It would be better to just ignore that invalid IP
and continue to check the other IPs in the list.

closes #38


ivarconr:
That would require a slight redesign of the strategy abstraction.

PR #69, opened 2018-01-27
Multi variant extensions

vsandvold:
This PR adds two methods to the client, needed for implementing an experimental multi variant strategy API.


elhoyos:
I understand this PR intends to add some support for multiple variations. Exposing the toggle object means you can use whatever the `dataFormatter#toNewFormat` returns. How do you plan to use it? Are strategies treated as variations?

When planning to do https://github.com/Unleash/unleash/pull/297 I discarded that option to make it evident that the way you allocate variations based on strategies. Maybe I'm missing something here?

PS: thank you for this great toggling framework.

PR #71, opened 2018-02-14
[wip] Support and operator and grouping of logic/strategies

sveisvei:
Work-in-progress to support `groups` and `AND` operator.

# Groups

(wip)

# Operator support

(wip)

# Caveats
* no precedence between AND/OR, order precedence only.

Format notes/spec:
```
a: (1 AND 2 AND 3) OR 4

    group: [
        {
            group: [{1, operator: 'AND'}, {2, operator: 'AND'}, {3}]
        },
        {
            4,
        }
    ]

b: (1 AND 2) OR 3 OR 4

    group: [
        {
            group: [{1, operator: 'AND'}, {2}]
        },
        {
            3,
        }
        {
            4,
        }
    ]


c: 1 OR 2 OR 3 OR 4

        [
            {1}, {2}, {3}, {4}
        ]


d: (1 OR 2) AND 3 AND 4

    group: [
        {
            group: [
                {1},
                {2}
            ],
            operator: 'AND'
        },
        {
            3,
            operator: 'AND'
        }
        {
            4,
        }
    ]

e: 1 OR (2 AND 3 AND 4)

    [
        {1},
        {
            group: [
                {2, operator: 'AND'},
                {3, operator: 'AND'},
                {4}
            ]
        }
    ]
```

### DIFF
--- a/src/strategy/strategy.ts
+++ b/src/strategy/strategy.ts
@@ -14,9 +14,14 @@ export class Strategy {
     }
 }
 
+enum StrategyGroupOperator {
+    AND,
+    OR
+}
+
 export interface StrategyTransportInterface {
     name: string;
     parameters: any;
-    operator?: string;
+    operator?: StrategyGroupOperator;
     group?: StrategyTransportInterface[];
 }
--- a/src/strategy/strategy.ts
+++ b/src/strategy/strategy.ts
@@ -17,4 +17,6 @@ export class Strategy {
 export interface StrategyTransportInterface {
     name: string;
     parameters: any;
+    operator?: string;
+    group?: StrategyTransportInterface[];
 }
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -196,3 +196,244 @@ test('should emit error when invalid feature runtime', t => {
 
     t.true(client.isEnabled('feature-wrong-strategy') === false);
 });
+
+function testGroup(group) {
+    const repo = {
+        getToggle() {
+            return buildToggle('feature', true, [
+                {
+                    name: 'unleash-internal-group-strategy',
+                    group,
+                },
+            ]);
+        },
+    };
+    const strategies = [new CustomFalseStrategy(), new CustomStrategy()];
+    const client = new Client(repo, strategies);
+    client.on('error', log).on('warn', log);
+    return client.isEnabled('feature');
+}
+
+test('should return the group value for sub lists when all is true', t => {
+    const result = testGroup([{ name: 'custom', operator: 'AND' }, { name: 'custom' }]);
+
+    t.true(result === true);
+
+    const result2 = testGroup([
+        { name: 'custom', operator: 'AND' },
+        { name: 'custom', operator: 'AND' },
+        { name: 'custom', operator: 'AND' },
+        { name: 'custom' },
+    ]);
+
+    t.true(result2 === true);
+
+    const result3 = testGroup([{ name: 'custom' }]);
+
+    t.true(result3 === true);
+});
+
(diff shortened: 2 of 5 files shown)


test/client.test.js
```diff
@@ -196,3 +196,244 @@ test('should emit error when invalid feature runtime', t => {
...
+
+    t.true(result3 === false);
+});
+
+test('groups with inner AND group that resovles to true', t => {
+    const result = testGroup([
+        {
+            name: 'inner-group',
+            group: [{ name: 'custom', operator: 'AND' }, { name: 'custom' }],
+            operator: 'AND',
```
ivarconr: Does it ever make sense to have `operator` defined on the group-level if the code only looks on it on strategy-definition level inside groups?

src/strategy/strategy.ts
```diff
@@ -17,4 +17,6 @@ export class Strategy {
 export interface StrategyTransportInterface {
     name: string;
     parameters: any;
+    operator?: string;
+    group?: StrategyTransportInterface[];
```
ivarconr: Would it make sense to define a group-interface that extends the `StrategyTransportInterface` instead?

```typescript
export interface StrategyGroupTransportInterface extends StrategyTransportInterface {
    operator: string;
    group: StrategyTransportInterface[];
}
```

src/strategy/strategy.ts
```diff
@@ -17,4 +17,6 @@ export class Strategy {
 export interface StrategyTransportInterface {
     name: string;
     parameters: any;
+    operator?: string;
```
ivarconr: Could the operator be an `Enum` type instead? 

```typescript
enum StrategyGroupOperator {
    AND,
    OR
}
```

ivarconr:
I think this looks promsing! 

I think there are a few typos in you spec, a), b) and d) the "outer" Array will never be a group? 

These should probably be:


**a: (1 AND 2 AND 3) OR 4**

```javascript
    [
        {
            group: [{1, operator: 'AND'}, {2, operator: 'AND'}, {3}]
        },
        {
            4,
        }
    ]
```

**b: (1 AND 2) OR 3 OR 4**

```javascript
    [
        {
            group: [{1, operator: 'AND'}, {2, operator: 'AND'}]
        },
        {
            3,
        }
        {
            4,
        }
    ]
```

**d: (1 OR 2) AND 3 AND 4**
```javascript
    [
        {
            group: [
                {1},
                {2}
            ],
            operator: 'AND'
        },
        {
            3,
            operator: 'AND'
        }
        {
            4,
        }
    ]
```

We should probably also require groups to have a `name` and `parameters` property. This will allow us to add the grouping capability, without breaking old clients, they will simply ignore groups. Remember that unleash promises that the unleash-api should always be one version backward compatible for clients, making it possible to upgrade unleash gradually.

src/strategy/strategy.ts
ivarconr: After revisiting the spec I realize that the operator should be on the `StrategyTransportInterface` and needs to be optional.

test/client.test.js
sveisvei: I think we initially should target to not use operator on the top-level, but further down the line I dont see any point on supporting legacy.

src/strategy/strategy.ts
sveisvei: 👍

src/strategy/strategy.ts
sveisvei: Indeed!

test/client.test.js
ivarconr: agreed :+1:

PR #73, opened 2018-02-27
Build on install

elhoyos:
I'm having trouble to `npm install` on a project using a fork of `unleash-client-node` because package.json is explicitly stating to never build on install.

#15 seems to be the reason behind it, but filesystem install tests suggests we don't need it anymore.


ivarconr:
@SimenB any objections?

SimenB:
I don't think so - I can't remember the issue.

elhoyos:
Please notice that despite this change works for a working directory installation for both `npm install` and `yarn`, it does not work for `yarn` when used as a github repository.

```
$ yarn add elhoyos/unleash-client-node#4a85ba6
$ ls -l node_modules/unleash-client/lib # empty
$ rm -rf node_modules
$ yarn
$ ls -l node_modules/unleash-client/lib # empty
```

PR #75, opened 2018-02-28
Variations support

elhoyos:
Here's a PoC supporting variations as per https://github.com/Unleash/unleash/pull/297

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,10 +1,18 @@
 import { EventEmitter } from 'events';
-import { Strategy, StrategyTransportInterface } from './strategy';
+import { Strategy, StrategyTransportInterface } from './strategy/index';
 import { FeatureInterface } from './feature';
+import { Variant } from './variant';
 import Repository from './repository';
+import { Experiment } from './strategy/strategy';
+import { experiment } from './index';
 
 interface BooleanMap {
     [key: string]: boolean;
+}
+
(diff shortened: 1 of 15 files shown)


ivarconr:
Awesome start! :100: 

I like that variants is put on as 

Two things that was not fully clear to me when looking through the code:

1. Is the user expected to provide their own `ExperimentStrategy` implementations? Would it make sense to provide a default implementation?
2. Will it make sense to have more than one `ExperimentStrategy` for a feature toggle? If yes, can you provide an example? (It's perfectly fine that the format support the possibility to have multiple experiment strategies, we don't have to expose this ability i the UI).

elhoyos:
> Is the user expected to provide their own ExperimentStrategy implementations?

Yes, if no strategy is provided `null` will be returned.

> Would it make sense to provide a default implementation?

I've changed my mind here while working with it compared to what I initially stated in Unleash/unleash#297. I do believe it could be very valuable for developers looking to understand what a basic implementation of an `Experiment` looks like.

> Will it make sense to have more than one `ExperimentStrategy` for a feature toggle?

🤔 Testing or business scenarios could benefit from multiple experiments:

```
[...strategies] AND
testing/marketing group: variant1, otherwise: null [targeted allocation experiment] OR
90%: control, 10%: variant1 [weighted allocation experiment]
```

PR #76, opened 2018-03-02
Build on prepare

elhoyos:
#73 fails to do the work for npm@^4.0.0 See [here](https://docs.npmjs.com/misc/scripts#deprecation-note).

```
$ npm -v
5.6.0
$ node -v
v6.9.5
```


SimenB:
Very nice! This is probably the behaviour I originally went for

PR #81, opened 2018-06-08
Make strategies optional in UnleashConfig

antonvasin:
Default value for `strategies` is `[]` in the constructor so we can omit this key when we just want to use default strategies

### DIFF
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -22,7 +22,7 @@ export interface UnleashConfig {
     metricsInterval?: number;
     disableMetrics?: boolean;
     backupPath?: string;
-    strategies: Strategy[];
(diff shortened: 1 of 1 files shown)


antonvasin:
Ok, will add it today or tomorrow

antonvasin:
Well, it’s really confusing. What this PR does is fix invalid interface, because constructor have `        strategies = []` already. Change to optional field only makes it identical to actual implementation of this interface. Also almost every test case in `unleash.test.js` instantiates new unleash client without explicit `strategies` prop. I really don’t think this PR needs another test

ivarconr:
Yes indeed, you are absolutely correct.

PR #101, opened 2019-01-24
[wip] Variants support

sveisvei:


src/client.ts
```diff
@@ -52,14 +54,22 @@ export default class UnleashClient extends EventEmitter {
     }
 
     isEnabled(name: string, context: any, fallbackValue?: boolean): boolean {
```
ivarconr: Note to self: Context should be a defined interface.

src/client.ts
```diff
@@ -69,25 +79,57 @@ export default class UnleashClient extends EventEmitter {
                     `Malformed feature, strategies not an array, is a ${typeof feature.strategies}`,
                 ),
             );
-            return false;
+            return { enabled: false };
```
ivarconr: Why does this method not simply return the `boolean` value?

src/client.ts
```diff
@@ -52,14 +54,22 @@ export default class UnleashClient extends EventEmitter {
...
 
     isEnabled(name: string, context: any, fallbackValue?: boolean): boolean {
+        const newContext: Context = {
+            ...context,
+            toggleName: name,
+        };
+        return this._isEnabled(name, newContext, fallbackValue).enabled;
+    }
+
+    _isEnabled(name: string, context: Context, fallbackValue?: boolean) {
```
ivarconr: can't this method the the `featureToggleDefinition` as input?

src/client.ts
```diff
@@ -69,25 +79,57 @@ export default class UnleashClient extends EventEmitter {
...
+        if (typeof feature === 'undefined' || !enabled) {
+            return fallbackVariant;
+        }
+
+        const variant: VariantInterface | null = selectVariant(feature, newContext);
+        if (variant === null) {
+            return fallbackVariant;
+        }
+
+        return {
```
ivarconr: This should be a defined interface called `Variant`

src/variant.ts
```diff
@@ -0,0 +1,93 @@
...
+    payload: PayloadInterface;
+}
+
+export interface VariantResult {
+    name: string;
+    payload: PayloadInterface;
+    enabled: boolean;
+}
+
+export function getDefaultVariantResult(): VariantResult {
```
ivarconr: I would like to simply call this `Variant`.

src/variant.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function selectVariant(
+    feature: FeatureInterface,
+    context: Context,
+): VariantInterface | null {
+    const userIdString = getUserFromContext(context);
+
+    if (!userIdString) {
+        return null;
```
ivarconr: We need to return a Variant even if there is no userId on the context. 

If the context is missing at all we probably need to just do some random things, this needs to be documented in the spec somehow.

src/client.ts
sveisvei: So it can be used by getVariant.

src/client.ts
sveisvei: Sure

src/client.ts
sveisvei: Changed the wrapper :)

src/variant.ts
sveisvei: done

src/variant.ts
sveisvei: done

test/client.test.js
```diff
@@ -181,3 +182,86 @@ test('should emit error when mising feature runtime', t => {
...
+        getToggle() {
+            return buildToggle('feature-but-no-variant', true, []);
+        },
+    };
+
+    const client = new Client(repo);
+
+    client.on('error', log).on('warn', log);
+    const result = client.getVariant('feature-but-no-variant', {});
+    const defeaultVariant = {
```
ivarconr: typo

src/index.ts
```diff
@@ -22,6 +23,17 @@ export function getFeatureToggleDefinition(toggleName: string) {
     return instance && instance.getFeatureToggleDefinition(toggleName);
 }
 
+export function getVariant(name: string, context: any, fallbackVariant?: Variant): Variant {
```
ivarconr: context should also be optional, can default to `{}`

batjko:
Awesome, can't wait to try this.
Thank you and nice work! :champagne:

PR #102, opened 2019-02-03
feat: Added support for variant override

ivarconr:


src/variant.ts
```diff
@@ -46,6 +52,16 @@ function getSeed(context: Context): string {
     return result || String(Math.round(Math.random() * 100000));
 }
 
+function overrideMatchesContext(context: Context): any {
```
ivarconr: @sveisvei Do you know of to define the return type here: "a Function which returns a boolean"

sveisvei:
```
function overrideMatchesContext(context: Context): (a: string) => boolean {
}
```

src/variant.ts
```diff
@@ -6,6 +6,11 @@ enum PayloadType {
     STRING = 'string',
 }
 
+interface Override {
+    field: string;
```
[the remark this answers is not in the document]
ivarconr: yes, I agree! 

This will need to be changed in frontend, server and specifications too. I will fix!

PR #108, opened 2019-04-16
Provide Browser support - #unleash in frontend

pPanda-beta:
### Purpose : 
Add support for browsers to this client so that it can be used in both node based backend as well as front apps
Right now available solution "Unleash Client React" does not support all features of unleash and that is only useable in react apps. 
On the other hand this client can be used in any javascript application. 

### How?

1. Added [localstorage-fs](https://www.npmjs.com/package/localstorage-fs) to replicate NodeJs fs on frontend and store toggles state in browser
2. Added [bfs-path](https://www.npmjs.com/package/bfs-path) to replicate NodeJs path library
3. Replaced [request](https://www.npmjs.com/package/request) library with [axios](https://www.npmjs.com/package/axios) since axios works with both browser and node js
4. Given compatible class for "NodeJs.Timer" interface

### What's next? 
Replace [localstorage-fs](https://www.npmjs.com/package/localstorage-fs) with [BrowserFS](https://www.npmjs.com/package/browserfs) so that a variety of storage medium (e.g. localstorage, indexed db,...) can be used. This can not be done right now as BrowserFs uses an older version of @types/node 

### Additional : 

- Duplicated test cases for Storage.ts to verify it works fine with localstorage implementation. 
- Tested with one react single page app using webpack

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ build/Release
 # typescript lives in src, output in lib
 lib/
 .vscode
+
+.idea
+**/*.iml
--- a/package-lock.json
+++ b/package-lock.json
@@ -861,6 +861,11 @@
         "tweetnacl": "^0.14.3"
       }
     },
+    "bfs-path": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/bfs-path/-/bfs-path-1.0.2.tgz",
+      "integrity": "sha512-KTKx2JJtAAAT7C/rJYDXXWA2VLPycAS4kwFktKsxUo0hj4UTtw/Gm5PJuY7Uf3xSlIQNo7HRCSWei2ivncVwbQ=="
+    },
     "binary-extensions": {
       "version": "1.13.0",
       "resolved": "https://registry.npmjs.org/binary-extensions/-/binary-extensions-1.13.0.tgz",
@@ -2187,6 +2192,11 @@
         "map-cache": "^0.2.2"
       }
     },
+    "fs-stats": {
+      "version": "0.0.0",
+      "resolved": "https://registry.npmjs.org/fs-stats/-/fs-stats-0.0.0.tgz",
+      "integrity": "sha1-i+clkYLwCU4pyD8JhD3JLEG3EWg="
+    },
     "fs.realpath": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
@@ -3111,8 +3121,7 @@
     "inherits": {
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
-      "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4=",
-      "dev": true
+      "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4="
     },
     "ini": {
       "version": "1.3.5",
@@ -3830,6 +3839,16 @@
         "parse-json": "^4.0.0",
         "pify": "^3.0.0",
         "strip-bom": "^3.0.0"
+      }
+    },
+    "localstorage-fs": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/localstorage-fs/-/localstorage-fs-0.1.0.tgz",
+      "integrity": "sha1-DaoTqtlzM5krboff1HzoSyiq0wI=",
+      "requires": {
+        "fs-stats": "~0.0.0",
+        "inherits": "~2.0.1",
+        "process": "~0.5.1"
       }
     },
     "locate-path": {
@@ -5872,6 +5891,11 @@
       "requires": {
         "parse-ms": "^2.0.0"
       }
+    },
+    "process": {
+      "version": "0.5.2",
+      "resolved": "https://registry.npmjs.org/process/-/process-0.5.2.tgz",
+      "integrity": "sha1-FjjYqONML0QKkduVq5rrZ3/Bhc8="
     },
     "process-nextick-args": {
       "version": "2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -407,21 +407,6 @@
         "any-observable": "^0.3.0"
       }
     },
-    "@types/caseless": {
-      "version": "0.12.2",
-      "resolved": "https://registry.npmjs.org/@types/caseless/-/caseless-0.12.2.tgz",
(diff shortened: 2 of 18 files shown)


ivarconr:
Hi @pPanda-beta, sorry for the slow response. 

Thanks for taking interest in the project and pushing for browser support. I realise that I probably have to put some effort in to this soon as users are asking for support directly in the browser. 

I am hover not sure if retrofitting the node-client in to also work in the browser is the right approach and I have discussed this in more detail in https://github.com/Unleash/unleash/issues/414

Would you mind reading it and give me your feedback on how to move this forward?

pPanda-beta:
Hi @ivarconr , sorry for the late reply. 
Thanks for guiding me to [Unleash/unleash#414](https://github.com/Unleash/unleash/issues/414). 
I've also looked into [Unleash/unleash#383](https://github.com/Unleash/unleash/issues/383).

What I could identify from #414 are mainly :

### 1. Huge traffic on unleash server :
- not really, most of the responses will be 304 with empty body
- If we choose the unleash-native-proxy to handle huge traffic then the it needs to be scaled, as an alternative we can scale the unleash server altogether. Introducing proxy is just a command-query segregation in the toggle service. 
- With #108 this problem still exists, and can only be solved by horizontally scaling  unleash-server and reducing the pull frequency 

### 2. Continuous pull which requires more resource : 
- [Unleash/unleash#383](https://github.com/Unleash/unleash/issues/383) can solve it by introducing poll / push based mechanism using SSE (js EventSource) or WebSocket
But it is more related with the sync mechanism not this node client's architecture 
- With #108 this can be solved by using proper pull frequency

### 3. Initial load problem :
- #108 already solves it by using localstorage to cache the last fetched values


But still what I would like to understand here is that these problems are mainly related with the client-server architecture of unleash. 
If someone wants to use the existing client in frontend then there is no readymade solution. 
Moreover with the unlease-native-proxy we need to create a separate client, or FINN team needs to open source it. 

Considering a small scale project it will be too much costly to spin up another proxy service for a little frontend userbase. On the other hand #108 just improves the existing client to be compatible with frontend. If these problems are really becoming a bottleneck then one can look for the alternative client+proxy solution.

ivarconr:
> not really, most of the responses will be 304 with empty body

My biggest **performance-concern** is that each request now does a query to DB even when returning 304. This can of course be fixed with various tactics. 

There is also the **security concern** where all toggles and their configuration (useIds etc) ends up being exposed to everyone. 

> Considering a small scale project it will be too much costly to spin up another proxy service for a little frontend userbase. On the other hand #108 just improves the existing client to be compatible with frontend. If these problems are really becoming a bottleneck then one can look for the alternative client+proxy solution.

Fair point. It might be sufficient in some use cases for small scale project to just make sure the this client is compatible. 

To learn more about this approach and how this can be used in a real app; maybe I can merge it in to a separate branch and make a beta release?

pPanda-beta:
> My biggest performance-concern is that each request now does a query to DB even when returning 304. This can of course be fixed with various tactics.

I totally agree with this. But I would suggest that this problem should be handled from unleash-server side. May be connecting redis or in-memory cache can solve it. (I'm assuming for most of the cases the total no of toggles, organization wide, **excluding metrics data**, will be very less) 

> There is also the security concern where all toggles and their configuration (useIds etc) ends up being exposed to everyone.

True, this can be fixed by requesting only a set of toggles from server.
#### But does this issue really applicable in every scenario?

1. **For a small scale project** only a few backend micro services will communicate with unleash-server, where adding this filter in api will be over engineering.
2. **For large scale project** / where the entire organization uses one toggle service to maintain toggles for all features of all business solutions, this could be useful. In this case one solution's micro service will never fetch toggles for another micro service's toggles. 
3. **For frontend clients** this will definitely be useful but in most of the industries people have not adopted the culture of micro-frontend. So mostly the frontend is combination of all solutions and their features. Hence it requires the toggles for everything. Which means the frontend needs all toggles so no requirement of security.

Example : 
Lets say Zuber is online cab service and have some micro services in backend.  May be zuber-payment, zuber-ride, zuber-cabs, zuber-customer-bff, zuber-driver-bff,... etc and has two frontend apps Zuber-customer and Zuber-driver.
Now Zuber-driver app may not require all the toggles of all those micro services, lets say zuber-payment. But Zuber-customer is the ultimate frontend app which requires toggles for most of the micro services. E.g. For a feature say "rating of driver", both Zuber-customer and Zuber-driver app requires the toggle info to enable it.
So the security filtering is useful for Zuber-driver app only. 


> To learn more about this approach and how this can be used in a real app; maybe I can merge it in to a separate branch and make a beta release?

It will be really nice if you can make a beta release or publish an npm package with different tag (may be unleash-client@frontend) so that I can try it in one of my current organization's project and share the feedback with you.

ivarconr:
@pPanda-beta finally made a tagged release: `unleash-client:3.2.3-frontend.0`. 

> so that I can try it in one of my current organization's project and share the feedback with you.

This will be highly appreciated. Questions you might answer:
- How did you integrate it in to your app (build steps etc)?
- what happens with metrics? Does it make sense to have all users sending metrics or is it better to disable that?
- Performance?

pPanda-beta:
Thanks @ivarconr 

> How did you integrate it in to your app (build steps etc)?


After building I've tested it with one of my organization's project. The project uses webpack for building.
Integration was pretty much straight forward.
1. npm i \<tarfile\> (now this can be `npm i unleash-client@3.2.3-frontend.0` )
2. Inside code I used it exactly the same way it is used for node based services.
 
Webpack versions were :
```
    "webpack": "^4.16.5",
    "webpack-cli": "^3.1.0",
    "webpack-concat-plugin": "^3.0.0",
    "webpack-dev-server": "^3.1.5",
    "webpack-logging-plugin": "^0.1.2",
    "webpack-version-file": "^0.1.3"
```

> what happens with metrics? Does it make sense to have all users sending metrics or is it better to disable that?
Performance?

This I can tell you after using it in that project on production. It will be prioritized in the upcoming sprints. I'll share the feedback after using it on prod for couple of weeks.

andreas1327250:
Is there any new progress on this? We are also thinking about using unleash and I also would prefer to just make the unleash-api publicly available and access it from the browser or our native apps - the proxy (as discussed in https://github.com/Unleash/unleash/issues/414) seems really like an unnecessary intermediate step.

ivarconr:
There is a browser friendly release you can try. I need some feedback on this before we adjust the official release

https://www.npmjs.com/package/unleash-client/v/3.2.3-frontend.0

tobias-jenkner:
Do you know when you will officially release this feature ? 

(We are currently using the 3.2.3-frontend.0 release.)

ivarconr:
The browser client is part of the unleash-hosted offering together with the [unleash-proxy](https://www.unleash-hosted.com/articles/the-unleash-proxy):

https://github.com/unleash-hosted/unleash-proxy-client-js

PR #115, opened 2019-05-08
feat: add timeout option for http requests

moolen:
I'd like to propose to add a `timeout` option to unleash in order to be able to control the HTTP request timeout. The current implementation has already the timeout configurable for `get()` and `post()`. However, this is not accessible for the user.

Why?
We're using Unleash in a (relatively) time-critical path and we'd like to be able to limit the time spent on upstream requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ The initialize method takes the following arguments:
 -   **strategies** - Custom activation strategies to be used.
 -   **disableMetrics** - disable metrics
 -   **customHeaders** - Provide a map(object) of custom headers to be sent to the unleash-server
+-   **timeout** - specify a timeout in milliseconds for outgoing HTTP requests. Defaults to 10000ms.
 
 ## Custom strategies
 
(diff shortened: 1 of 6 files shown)


moolen:
i dug into the tests because the coverage threshold was too low;
I figured out that `nyc` does not properly calculate the coverage for some `nock` based tests. I added annotations to "fix" that. Tell me what you think.

ivarconr:
You are aware the architecture of Unleash? The client SDK's works on local state and will update the local state in the background? For a `isEnabled` call the SDK will not to a http call at all. So the performance of the client>=>server communication should not affect the client application. 

But I do agree that it makes sense to introduce a way to set the http timeout.

Would you mind add some details about this parameter in the readme and specify that it expects ms precision?

moolen:
Thanks for your feedback @ivarconr! I added documentation. However, the integration test fails for this commit only :/. I'll dig into the test. there might be a race condition.

I'll open a issue for our use-case and describe it in more detail (that's not really related to this PR)

moolen:
I wasn't able to reproduce this issue locally :/

ivarconr:
Thanks, I will look in to this tomorrow.

PR #117, opened 2019-05-13
feature: inject repository

moolen:
Here's the PR to inject the repository. PTAL when you have time. 

What's missing from the docs:
* repository events: `data, ready, error`
* metrics events: `warn, error, registered, sent`
* storage events: `ready, persisted, error`
* client events: `warn, error`

I wasn't sure whether or not a user of this SDK should depend on the implementation details of the `repository`, `storage`, `metrics` and `client` objects. That makes changing the underlying implementation really hard. That's why i would discourage documenting that.
Though, it certainly is possible to access the internals and to subscribe to those events.
Please let me know if you want to have that documented, too. I'll add them in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,8 +155,38 @@ instance.on('ready', console.log.bind(console, 'ready'));
 instance.on('error', console.error);
 ```
 
+## Events
+
+The unleash instance object implements the EventEmitter class and **emits** the following events:
+
+| event      | payload                          | description                                                                              |
+| ---------- | -------------------------------- | ---------------------------------------------------------------------------------------- |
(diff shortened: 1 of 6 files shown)


moolen:
build fails due to dependency update of `husky@2.0.2` which dropped support for `node v6` which went EOL this month.

ivarconr:
Thanks, this looks really good.  Sorry for the delay, my week has been flying away and this ended to be on the bottom of things to follow up. 

> I wasn't sure whether or not a user of this SDK should depend on the implementation details of the repository, storage, metrics and client objects. That makes changing the underlying implementation really hard. That's why i would discourage documenting that. Though, it certainly is possible to access the internals and to subscribe to those events.

You have a good point .Let's postpone that to later.

PR #123, opened 2019-05-16
Feat: Get all feature toggle definitions at once 

sbuckpesch:
As we have different feature toggles for different teams on the same server instance it is important to us to request all available features toggles, filter them and pass the subset of toggles to our frontend clients.

The PR provides a method `getFeatureToggleDefinitions()` to the Unleash client including docs and tests.


sbuckpesch:
Same issue with Node v6 (#118).

ivarconr:
Thanks, it makes sense to me to add this capability. Would you mind rebase it on top of latest master?

sbuckpesch:
I merged the latest changes in my branch.

PR #126, opened 2019-05-20
fix: update README init param defaults to milliseconds

sbuckpesch:
metricsInterval and refreshInterval are milliseconds values, so add defaults to the docs in milliseconds as well.


elhoyos:
This makes a lot of sense @sbuckpesch. Thank you.

Issue #131, opened 2019-06-12
gradualRolloutRandom wrong with its implementation


zjy01:
`Math.round(Math.random() * 100) + 1` means `[1,101]`。
I think it should be `Math.floor(Math.random() * 100) + 1`, means `[1,100]`

PR #132, opened 2019-06-12
fix: Strategy gradualRolloutRandom random should <= 100

zjy01:
Fix for #131 
When I set parameters for gradualRolloutRandom as `100%`，it means always enabled.

### DIFF
--- a/test/strategy/gradual-rollout-random.test.js
+++ b/test/strategy/gradual-rollout-random.test.js
@@ -36,3 +36,9 @@ test('gradual-rollout-random strategy should be disabled when percentage=0', t =
     let params = { percentage: '0', groupId: 'test' };
     t.false(strategy.isEnabled(params));
 });
+
+test('gradual-rollout-random strategy should be enabled when percentage=100', t => {
+    const strategy = new GradualRolloutRandomStrategy();
+    let params = { percentage: '100', groupId: 'test' };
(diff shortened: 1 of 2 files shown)


elhoyos:
Nice catch @zjy01, thanks for putting this fix.

test/strategy/gradual-rollout-random.test.js
```diff
@@ -36,3 +36,9 @@ test('gradual-rollout-random strategy should be disabled when percentage=0', t =
     let params = { percentage: '0', groupId: 'test' };
     t.false(strategy.isEnabled(params));
 });
+
+test('gradual-rollout-random strategy should be enabled when percentage=100', t => {
```
ivarconr: This test is kind of poor. We should control random-ness in the test to make sure we always test the edge cases

elhoyos: I find useful the test added by @zjy01 here. But it certainly is not going for the edge cases.

How do you guys usually test that? [proxyquire](https://github.com/thlorenz/proxyquire) is how I would go.

ivarconr: I usually just add the possibility to inject a random-generator-function in the constructor which defaults to `Math.random`. In the test you just provide a function which produces the number you want to test.

PR #133, opened 2019-06-12
chore: control gradual rollout strategy randomness

elhoyos:
This is a follow up of https://github.com/Unleash/unleash-client-node/pull/132#discussion_r292938714.

Side note: do you see any advantage to exit early if percentage is 0?


ivarconr:
> Side note: do you see any advantage to exit early if percentage is 0? 

No.

PR #136, opened 2019-07-01
feat: Add support for static context fields

### LINKED ISSUE #135: feat: add support static context fields
In order to prepare the SDKS for [strategy constraints](https://github.com/Unleash/unleash/issues/442) we need to introduce two new fields in the [Unleash Context](https://unleash.github.io/docs/unleash_context) 

- `environment` (default value is "default")
- `appName` (already a required config parameter)

Both these static context fields should be configured at initialisation of the unleash SDK and be set as default values of the Unleash Context provided to the strategy implementations. The user is allowed to override these values in the context object at runtime (this is required to be backward compatible). 

This support should be possible to add without breaking any contract and should be released as a minor version (3.3.x).

ivarconr:
Added support for two new static context fields:
- environment
- appName

These are set in the Configuration when the client is
initialized and should be inserted in to the UnleashContext.

closes #135

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -3,5 +3,7 @@ export interface Context {
     userId?: string;
     sessionId?: string;
     remoteAddress?: string;
+    environment?: string;
+    appName?: string;
     properties?: any;
 }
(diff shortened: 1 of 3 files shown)


ivarconr:
TODO: needs some unit tests.

PR #152, opened 2019-11-02
feat: customHeaderFunction to dynamic setting headers

SecretBase:
Resolve https://github.com/Unleash/unleash-client-node/issues/151

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from 'events';
 import { Response } from 'request';
 import { resolve } from 'url';
 import { post, Data } from './request';
-import { CustomHeaders } from './unleash';
+import { CustomHeaders, CustomHeadersFunction } from './unleash';
 import { sdkVersion } from './details.json';
 
 export interface MetricsOptions {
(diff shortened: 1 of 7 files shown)


src/metrics.ts
```diff
@@ -126,7 +135,7 @@ export default class Metrics extends EventEmitter {
         return true;
     }
 
-    sendMetrics(): boolean {
+    async sendMetrics(): Promise<boolean> {
```
ivarconr: Making this return a Promsie would that not require the calling method to `await` it to make sure it actually is executed?  

(`this.sendMetrics();` on line 79).

src/metrics.ts
```diff
@@ -93,19 +97,24 @@ export default class Metrics extends EventEmitter {
         this.disabled = true;
     }
 
-    registerInstance(): boolean {
+    async registerInstance(): Promise<boolean> {
```
ivarconr: calls to this method should now be `awaiting` the response.

ivarconr: nvm; this is not a requirement i javascript

src/metrics.ts
ivarconr: nvm; this is not a requirement i javascript

ivarconr:
released as part of 3.3.0

PR #162, opened 2020-01-23
Fix sinon 7.4.0 reference in package-lock.json

mfolkeseth:
Sinon 7.4.0 is referenced in `package-lock.json` but is no longer present in the npmjs registry.

This causes `npm install` to fail completely.

The fix is to remove `package-lock.json` and make a new one with `npm install`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,358 +5,214 @@
   "requires": true,
   "dependencies": {
     "@ava/babel-plugin-throws-helper": {
-      "version": "3.0.0",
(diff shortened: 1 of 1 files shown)


ivarconr:
this project uses yarn and not npm for lock files, it is a mistake that package-lock.json was checked in.

ivarconr:
https://github.com/Unleash/unleash-client-node/commit/7f7225f93b3f5dd6158aed37e9ed7408ff5868c2

PR #164, opened 2020-01-23
Emit new events

mfolkeseth:
Emit events when
* Request has successfully been made to server `serveralive`
* Changes to toggles has been made `repochanges`

`serveralive` is very useful to register that the client is reconnected to server.

https://github.com/Unleash/unleash-client-node/issues/163

### DIFF
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -133,6 +133,14 @@ export class Unleash extends EventEmitter {
             this.emit('warn', msg);
         });
 
+        this.repository.on('serveralive', () => {
+            this.emit('serveralive');
+        });
+
+        this.repository.on('repochange', data => {
+            this.emit('repochange', data);
+        });
+
         this.metrics = new Metrics({
             disableMetrics,
             appName,
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -133,12 +133,12 @@ export class Unleash extends EventEmitter {
             this.emit('warn', msg);
         });
 
-        this.repository.on('serveralive', () => {
-            this.emit('serveralive');
+        this.repository.on('unchanged', () => {
+            this.emit('unchanged');
         });
 
-        this.repository.on('repochange', data => {
-            this.emit('repochange', data);
+        this.repository.on('changed', data => {
+            this.emit('changed', data);
         });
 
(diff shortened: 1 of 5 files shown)


src/unleash.ts
```diff
@@ -133,6 +133,14 @@ export class Unleash extends EventEmitter {
             this.emit('warn', msg);
         });
 
+        this.repository.on('serveralive', () => {
+            this.emit('serveralive');
+        });
+
+        this.repository.on('repochange', data => {
+            this.emit('repochange', data);
```
ivarconr: I think i would prefer the event name to simply be `update`.

mfolkeseth: Agreed. I will change it tomorrow morning. In retrospect I am not too keen on the”serveralive” either. It is not what it represents. How about “responding”, or  “connected”?

ivarconr: I was thinking the same thing, but I did not have any better suggestion. What do you intend to use serveralive for?

What about: 
- `changed` - when toggle configuration are updated
- `unchanged` - 304 response from server, nothing new. 

(The Java client just returns a FearureToggle response with the status code as data. I'm not convinced that's any better.)

mfolkeseth: I like that. Let us go with `changed` and `unchanged`.

The intended use for `serveralive` is to let the client know if it was able to reach the server. By combining `changed` and `unchanged` you would get the same effect.

So now you can control what to do if unleash crashes or something else goes wrong. In this case, fallback to fa falsy response. That is kinda cool:
```
let connected = false
const instance = initialize({...})

instance.on('unchanged', () => {
  connected = true
})

instance.on('changed', () => {
  connected = true
})

instance.on('error', () => {
  connected = false
})

function isEnabledWithFallback(toggle, context, fallback) {
  if (!connected) { return false }  // fallback to false if connection to unleash somehow goes wrong
  return isEnabled(toggle, context, fallback)
}

module.exports = {
  isEnabled: isEnabledWithFallback
} 
```

ivarconr:
Had to do a manual merge: https://github.com/Unleash/unleash-client-node/commit/d9a0f84a9d4057fbd75de1538c0af04f2da5e539

PR #185, opened 2020-09-02
feat: Add stale property to the FeatureInterface

vtremblay:
We need to be able to see if a feature flag is marked "deprecated" during runtime in order to act accordingly

### DIFF
--- a/src/feature.ts
+++ b/src/feature.ts
@@ -5,6 +5,7 @@ export interface FeatureInterface {
     name: string;
     description?: string;
     enabled: boolean;
+    stale: boolean;
     strategies: StrategyTransportInterface[];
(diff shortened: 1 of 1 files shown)


ivarconr:
Great, you are already disclosing our secret plan of introducing this property! The idea is to add feedback mechanisms when the application is using a stale feature flag. 

Can you, @vtremblay, please elaborate a bit on how you plan to use this information in your set-up?

vtremblay:
@ivarconr We have a small wrapper around the SDK to help us manage the dependency. That plan, for now, was to log a warning or an error each time a stale flag is being evaluated. We will probably come with something more elegant in the future, but that a starting point!

PR #188, opened 2020-09-06
fix: Only unref timers when used in node, where unref is defined

### LINKED ISSUE #187: TypeError: this.timer.unref is not a function
I get this error `TypeError: this.timer.unref is not a function` when using this library in Electron, within the render process. I think that it is because there is a difference in a timing function implementation (maybe setInterval?) between node and the frontend renderer. Is this possible to solve? I think that this library should work in Electron besides this, since node is available, it just for some reason uses the wrong timing function here.

alfredvaa:
Resolves #187. The library could not be used in Electron because Electron seems to not use `setTimeout` from Node, and therefore the `unref` method does not exist. The library seems to be working without that method when running inside a render process, so this might be a solution. Not sure about if this is a solid solution, how about it?

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -83,7 +83,7 @@ export default class Metrics extends EventEmitter {
             this.sendMetrics();
         }, this.metricsInterval);
 
-        if (process.env.NODE_ENV !== 'test') {
+        if (process.env.NODE_ENV !== 'test' && typeof this.timer.unref === 'function') {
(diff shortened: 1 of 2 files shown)


sarendsen:
Would be nice to see this merged.

sveisvei (approved):
Looks good to me.

ivarconr:
thanks for the push, released as part of `3.4.0`

PR #197, opened 2020-12-08
feat: use isomorphic-fetch

menduz:
Hello my friends, I really wanted to use unleash inside a Cloudflare Worker to apply feature flags on edge workers to serve different versions of our apps. But there is no `request` module in that environment. So I adapted your library to use `isomorphic-fetch` that would enable this lib to run inside service workers, cloudflare workers, node, deno and the browser.

I won't feel bad if you do not merge this PR since it may not be in your plans or roadmap. But I feel obligated to contribute with a PR to give it a chance.


src/request.ts
```diff
@@ -23,54 +22,40 @@ export interface PostRequestOptions extends RequestOptions {
...
-) => {
-    const options = {
-        url,
-        timeout: timeout || 10000,
-        agentOptions: {
-            keepAlive: true,
-            keepAliveMsecs: 30 * 1000,
-            timeout: timeout || 10 * 1000,
-        },
+export const get = async ({ url, etag, appName, instanceId, headers }: GetRequestOptions) => {
```
ivarconr, line 49: It is really important that we use proper [agent settings](https://nodejs.org/api/http.html) to leverage correct keep-alive for the SDK. Without it the SDK will create new connections for every request and reuse global keepAlive settings.

ivarconr:
Thanks for showing the way. 

I was initially thinking to switch to node-fetch ([see issue](https://github.com/Unleash/unleash-client-node/issues/169#issuecomment-697225523)). 

Personally I do not fancy that `isomorphic-fetch` establishing `fetch` as a global variable. My initial idea was that the SDK should use "node-fetch" by default, but allow  the use to inject a custom fetch implementation as part of the configuration. I'm not sure if that would approach would work for your Cloudflare Worker scenario?

ivarconr:
@menduz what do you think about using https://github.com/lquixada/cross-fetch instead?

ivarconr:
Hi again, 

I am really sorry for leaving this around for a while. I have been busy doing other things, but we are now ready to focus more on this SDK again, full speed ahead. 

We just changed to using node-fetch, and this is already part of the main branch. 

I would not have the SDK depend directly on `isomorphic-fetch`, but I am willing to consider a [browser specific](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser) entry for the browser in the `package.json`. 

The browser enrty needs to do a couple of things:
- change the backup-handler to use localstorage instead of file based storage. 
- create a specific "request" module for the browser context (e.g. http(s) agents does not make sense in the browser, and fetch is available in the global scope)

SimenB:
If you're adding a `browser` field to package.json, you should also consider `exports`.

- https://nodejs.org/api/packages.html#packages_conditional_exports
- https://webpack.js.org/guides/package-exports/

ivarconr:
Closing this, but created a separate issue to start the discussions on how we can add support for browsers as well with this SDK: #213

PR #201, opened 2021-01-27
feat: flexible rollout - custom stickiness

sbelzile-nexapp:
Stickiness on the flexible rollout strategy could be more generic.

The use case that I have:

- Our users are always part of an "organization"
- There are features that we want to rollout for all users of, let's say, 50% of organizations.

What I did:

- I added an additional stickiness type: `custom`.
- When `custom` is specified, a `contextField` must be specified as well.
- When `custom` is specified, we base stickiness based on the value of the `contextField` on the provided `context`.
- When `custom` is specified and `contextField` or it's value in the `context` is null or undefined, we return false and the flag is disabled.

### DIFF
--- a/src/strategy/flexible-rollout-strategy.ts
+++ b/src/strategy/flexible-rollout-strategy.ts
@@ -33,7 +33,6 @@ export class FlexibleRolloutStrategy extends Strategy {
         const groupId = parameters.groupId || context.featureToggle || '';
         const percentage = Number(parameters.rollout);
         const stickiness: string = parameters.stickiness || STICKINESS.default;
-
         const stickinessId = this.resolveStickiness(stickiness, context);
 
         if (!stickinessId) {
--- a/src/strategy/flexible-rollout-strategy.ts
+++ b/src/strategy/flexible-rollout-strategy.ts
@@ -4,6 +4,7 @@ import { normalizedValue } from './util';
 
 const STICKINESS = {
     default: 'default',
+    custom: 'custom',
     userId: 'userId',
     sessionId: 'sessionId',
     random: 'random',
@@ -19,12 +20,14 @@ export class FlexibleRolloutStrategy extends Strategy {
         }
     }
 
-    resolveStickiness(stickiness: string, context: Context): any {
+    resolveStickiness(stickiness: string, context: Context, customField?: string): any {
         switch (stickiness) {
             case STICKINESS.userId:
                 return context.userId;
             case STICKINESS.sessionId:
                 return context.sessionId;
+            case STICKINESS.custom:
+                return customField ? context[customField] : null;
             case STICKINESS.random:
                 return this.randomGenerator();
             default:
@@ -36,7 +39,7 @@ export class FlexibleRolloutStrategy extends Strategy {
         const groupId = parameters.groupId || context.featureToggle || '';
         const percentage = Number(parameters.rollout);
         const stickiness = parameters.stickiness || STICKINESS.default;
-        const stickinessId = this.resolveStickiness(stickiness, context);
+        const stickinessId = this.resolveStickiness(stickiness, context, parameters.customField);
 
         if (!stickinessId) {
             return false;
--- a/src/strategy/flexible-rollout-strategy.ts
+++ b/src/strategy/flexible-rollout-strategy.ts
@@ -33,6 +33,7 @@ export class FlexibleRolloutStrategy extends Strategy {
         const groupId = parameters.groupId || context.featureToggle || '';
         const percentage = Number(parameters.rollout);
         const stickiness: string = parameters.stickiness || STICKINESS.default;
+
         const stickinessId = this.resolveStickiness(stickiness, context);
 
         if (!stickinessId) {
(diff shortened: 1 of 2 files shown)


ivarconr:
Hi @sbelzile-nexapp,

Thanks for your suggestions. It has always been our intention that you should be able to also be sticky on custom context fields. 

What I have been thinking is that the "stickiness dropdown" in the UI could be populated with the defined context fields, introduced as part of the "strategy constraint" feature (enterprise). 

image

this would require us to move the ability to define custom context fields to open-source, and it would then require use to change in the SDKs to resolve the correct context field based on the actual name.

What do you think?

sbelzile-nexapp:
So, in there, I could just replace `custom` "type" directly by my `contextField` value.

It makes sense and would simplify the code.

ivarconr:
> So, in there, I could just replace custom "type" directly by my contextField value.

Exactly, so an example strategy configuration would the be:

```json
{
	"name": "flexibleRollout",
	"parameters": {
		"rollout": "100",
		"stickiness": "organisation",
		"groupId": "Demo"
	}
}
```

and the user would then have to provide the "organisation" on the unleash context.

src/strategy/flexible-rollout-strategy.ts
```diff
@@ -19,16 +17,14 @@ export class FlexibleRolloutStrategy extends Strategy {
...
-                return context.userId;
-            case STICKINESS.sessionId:
-                return context.sessionId;
+            case STICKINESS.default:
+                return context.userId || context.sessionId || this.randomGenerator();
             case STICKINESS.random:
                 return this.randomGenerator();
             default:
-                return context.userId || context.sessionId || this.randomGenerator();
+                return stickiness ? context[stickiness] : null;
```
ivarconr, line 27: Hi, This almost correct. Because a poor decision in the early days, pre typescript, the context is a bit ambiguous when it comes to custom context fields. The initial idea was that all "not official defined context fields" should be defined under `context.properties` ([source code](https://github.com/Unleash/unleash-client-node/blob/master/src/context.ts#L12)). 

This means that we need to first check "context[stickiness]" and if it is undefined we also need to look inside "context.properties[stickiness]" (and context.properties may also be undefined).

ivarconr: There is actually a helper for this:
https://github.com/Unleash/unleash-client-node/blob/master/src/helpers.ts#L19

ivarconr:
I have started looking in to the GUI and API side of things. I think we will add a new field to the defined context fields where you may specify whether it should be possible to use this field for stickiness.

sbelzile-nexapp:
> I have started looking in to the GUI and API side of things. I think we will add a new field to the defined context fields where you may specify whether it should be possible to use this field for stickiness.

Not sure I understand what you mean by `new field to the defined context`: a field on the context object?

ivarconr:
> Not sure I understand what you mean by new field to the defined context: a field on the context object?

The idea is that you have to configure whether a "defined context field" should be allowed to be used as a criteria for stickiness. This to avoid the "stickiness list" on the "flexible roll-out strategy" becoming too long.

src/strategy/flexible-rollout-strategy.ts
```diff
@@ -19,16 +18,14 @@ export class FlexibleRolloutStrategy extends Strategy {
...
-                return context.userId;
-            case STICKINESS.sessionId:
-                return context.sessionId;
+            case STICKINESS.default:
+                return context.userId || context.sessionId || this.randomGenerator();
             case STICKINESS.random:
                 return this.randomGenerator();
             default:
-                return context.userId || context.sessionId || this.randomGenerator();
+                return stickiness ? resolveContextValue(context, stickiness) : null;
```
ivarconr, line 28: Would typescript not allow you to do just a `resolveContextValue(context, stickiness)` here? Is there any scenarios where stickiness can be undefined? If that's the case we should probably use the default scenario.

ivarconr:
I am in favor of this PR! :+1: 

I would like to look in the server/frontend side of things as well and have this part of a future version soon.

ivarconr (approved):

ivarconr:
we also need to write a few specifications for this changes in the built in flexible-rollout-strategy as well:  https://github.com/Unleash/client-specification

ivarconr:
I did a small PoC on the API and frontend. 

The flexibleRollout Strategy:
Screenshot

The context edit:
context_with_stickiness

src/strategy/flexible-rollout-strategy.ts
sbelzile-nexapp: You are right. I changed the method (`resolveStickiness`) signature to take a `string` instead of `string | undefined`. This allowed me to make that change.

ivarconr:
@chriswk do you see any reason not to merge this in now? 

It does not seem to break any existing functionality. It will not be very use-full before one also have support from the Unleash Admin UI, but one can start using this functionality by defining the toggle directly via the API.

chriswk:
No, I can't see a reason not merge it

chriswk (approved):
LGTM

PR #202, opened 2021-01-27
feat: variant stickiness

sbelzile-nexapp:
Same as https://github.com/Unleash/unleash-client-node/pull/201 but for variants

### DIFF
--- a/src/variant.ts
+++ b/src/variant.ts
@@ -21,6 +21,7 @@ export interface Payload {
 export interface VariantDefinition {
   name: string;
   weight: number;
+  stickiness?: string;
   payload: Payload;
   overrides: Override[];
 }
@@ -38,10 +39,18 @@ export function getDefaultVariant(): Variant {
   };
 }
 
-const stickynessSelectors = ['userId', 'sessionId', 'remoteAddress'];
-function getSeed(context: Context): string {
+function randomString() {
+  return String(Math.round(Math.random() * 100000));
+}
+
+const stickinessSelectors = ['userId', 'sessionId', 'remoteAddress'];
+function getSeed(context: Context, stickiness: string = 'default'): string {
+  if (stickiness !== 'default') {
+    const value = resolveContextValue(context, stickiness);
+    return value ? value.toString() : randomString();
+  }
   let result;
-  stickynessSelectors.some((key: string): boolean => {
+  stickinessSelectors.some((key: string): boolean => {
     const value = context[key];
     if (typeof value === 'string' && value !== '') {
       result = value;
@@ -49,7 +58,7 @@ function getSeed(context: Context): string {
     }
     return false;
   });
-  return result || String(Math.round(Math.random() * 100000));
+  return result || randomString();
 }
 
 function overrideMatchesContext(context: Context): (o: Override) => boolean {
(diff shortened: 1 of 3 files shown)


src/feature.ts
```diff
@@ -8,4 +8,5 @@ export interface FeatureInterface {
     stale: boolean;
     strategies: StrategyTransportInterface[];
     variants: VariantDefinition[];
+    variantStickiness?: string;
```
ivarconr, line 11: I am not too happy about introducing a new field to the feature toggle object, as it clutters the [data return from the api](https://unleash.github.io/docs/api/admin/features#fetching-feature-toggles) a bit. 

In a perfect world the variants would not be an array, but an object, so that we could attach the meta-config directly.

Maybe we can introduce the stickiness information on the variant objects inside the array? 

```json
{
   "name": "variant1",
   "weight": 50,
   "stickiness": "some-custom-field"
},
{
   "name": "variant2",
   "weight": 50,
   "stickiness": "some-custom-field"
}
````

This is not 100% optimal either, as you can not have different variants with different stickiness...

sbelzile-nexapp: > In a perfect world the variants would not be an array, but an object, so that we could attach the meta-config directly.

Yes, I think so as well. I did not want to introduce breaking changes.

I am not really convinced by the array either. Could somebody ever want to define different stickiness?

ivarconr: > Could somebody ever want to define different stickiness?

Good question. I'm not sure, but I think the override is more relevant if you need to control who sees what.

ivarconr:
I am a bit unsure of the value it brings to have this on the variants. If you require to control the stickiness on the variant level, would you then not rather go with the `variant overrides` capability instead and specify which org gets which variant?

sbelzile-nexapp:
> I am a bit unsure of the value it brings to have this on the variants. If you require to control the stickiness on the variant level, would you then not rather go with the variant overrides capability instead and specify which org gets which variant?

Correct me if I am wrong, I might not understand the override properly, but isn't the variant override a list of specific value?

My end goal is to get uniformity between my organizations: let's say 1/3 of my organizations get a blue header, another 1/3 get  a red header, and the final third get it yellow. I would like my users to get the same header everybody else in their organization gets.

If I use the variant override, I must specify exactly which 1/3 gets what, right? 1 to 5 organization is not that bad to specify manually, but 100+ will be a pain...

Don't you think having something that does the split automatically would be great?

ivarconr:
> If I use the variant override, I must specify exactly which 1/3 gets what, right? 1 to 5 organization is not that bad to specify manually, but 100+ will be a pain...

> Don't you think having something that does the split automatically would be great?

It makes sense in this scenario. My concern was more that we introduce extra options of the users. I think we can find ways without overloading the user with options.

ivarconr:
I have created a UI for updating the stickiness for variants as well. 

image


PS! We plan to ship customisable stickiness for flexibleRollout and variants in next feature release of Unleash.

ivarconr:
Hi again @sbelzile-nexapp, 

sorry it took me some time to have a review of this. I have updated the PR with a few simplifications and some more tests for my own sanity. 

fyi:  We have already released the necessary api's (part of unleash 3.12.x) and will include the updated ui as part if the next feature release. We plan to test this a bit under the covers in the node sdk before we bring similar support to the other SDKs.

ivarconr (approved):

PR #206, opened 2021-02-08
chore: Bump all dev-dependencies

chriswk:
- Migrating from eslint-config-finn to eslint-config-airbnb
  and enabling airbnb typescript as well.
- Fixes needed to get this to run without warnings.
  There is one dependency cycle between variants and feature that I
  couldn't find a way to untangle for now.
- AirBnb also fails on more than one class per file, so extracted some
  of our test classes to separate files.
- Run eslint . -fix
- Upgrade ava, and tell ava about our project configs of jsconfig and
  tsconfig.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,15 +5,9 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 4
+indent_size = 2
 insert_final_newline = true
 trim_trailing_whitespace = true
 
 [*.md]
 trim_trailing_whitespace = false
-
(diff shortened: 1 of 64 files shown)


SimenB:
@chriswk hah, didn't notice this 😀 I opened up #207 which updates ava and includes fixes for the node 10 test faiures

chriswk:
> @chriswk hah, didn't notice this  I opened up #207 which updates ava and includes fixes for the node 10 test failures

Thanks. I'll have a look at using that.

PR #207, opened 2021-02-10
chore: update to ava@3

SimenB:
Should in theory clear out the dependabot warnings in the repo


package.json
```diff
@@ -90,24 +91,15 @@
...
   "lint-staged": {
-    "*.js": [
-      "eslint --fix",
-      "git add"
-    ],
-    "*.{ts,md,json}": [
-      "prettier --write",
-      "git add"
-    ]
+    "*.js": "eslint --fix",
```
SimenB, line 94: image

package.json
```diff
@@ -41,13 +41,13 @@
     "examples"
   ],
   "devDependencies": {
+    "@ava/babel": "^1.0.1",
     "@types/ip": "^1.1.0",
     "@types/murmurhash3js": "^3.0.2",
-    "@types/nock": "^11.1.0",
```
SimenB, line 46: `nock` ships with types

SimenB:
Interestingly, this gave 3 test failures on node 10. Seems to be some sort of race condition. The tests looked somewhat faulty to me either way - e.g. multiple calls to `end`.

I added `t.plan` which made the tests fail with "wrong number of assertions" rather than "test already done", then fixed the tests (I think)

chriswk (approved):
This looks good. We should probably merge this before looking at my bigger upgrade all dependencies PR:#206 . And then rebase the #206 branch on top of master after merging this, reducing the surface that touches.

chriswk:
I updated #206 to include what you've done here :)

ivarconr (approved):

PR #208, opened 2021-02-10
chore: Clarify ready event in readme

ivarconr:


chriswk (approved):
LGTM

README.md
```diff
@@ -158,30 +162,30 @@ Its also possible to ship the unleash instance around yourself, instead of using
...
-| unchanged  | -                                | is emitted each time the client gets new toggle state from server, but nothing has changed  |
-| changed    | `object` data                    | is emitted each time the client gets new toggle state from server and changes has been made |
+| event      | payload                          | description                                                                                                                                                                                                                                  |
+| ---------- | -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ready      | -                                | is emitted once the fs-cache is ready. if no cache file exists it will still be emitted. The client is ready to use, but might not have synchronized with the Unleash API yet. This means the SDK still can operate on stale configurations. |
```
SimenB, line 181: does this mean it's `registered` we actually want to wait for if we want to wait for updated data?

ivarconr: No, registered only means the SDK has sent meta info to the API. 
For that use case you would wait for the first "changed" event today. 

We plan to add another event to make this clearer: https://github.com/Unleash/unleash-client-node/issues/209

SimenB: ah perfect, thanks!

Issue #179, opened 2020-06-16
If the custom headers functions fails, further fetches are not queued


AndreasChristianson:
If the [custom headers function](https://github.com/Unleash/unleash-client-node/blob/master/src/repository.ts#L107) rejects (for example during temporary network outage when refreshing a jwt), the [timed fetch](https://github.com/Unleash/unleash-client-node/blob/master/src/repository.ts#L121) never requeues. This leaves the client in a state where it is no longer attempting to refresh its cache from the unleash server.

This can be demonstrated by providing a [custom headers function](https://github.com/Unleash/unleash-client-node/blob/master/src/unleash.ts#L33) that fails every other call:

```
let callCount = 0;
const instance = initialize({
    url: 'http://unleash.herokuapp.com/api/',
    appName: 'my-app-name',
    instanceId: 'my-unique-instance-id',
    customHeaderFunction: () => new Promise((resolve, reject) => {
        if(++callCount % 2 === 0){
            reject();
        } else {
            resolve({
                myAuthHeader: 'secrets...'
            })
        }
    })
});
```
Using the above config, note that the client stops updating after the first failed refresh.

Potential solutions:
 - wrap the custom headers function call in a try/catch
 - move the `this.timedFetch()` above any user provided function calls

ivarconr:
thanks for reporting.

PR #210, opened 2021-02-10
fix: Keep fetching if customHeadersFunction fails.

ivarconr:
Moving the customHeadersFunction evalution inside the try/catch
and ensuring we do the timedFetch inside the finally block. This
guarantees that we start a new fetch even if the customHeadersFunction
throws an error.

fixes: #179


chriswk (approved):
Seems reasonable. It does seem like we'd like to continue going even if we couldn't build customHeaders

PR #211, opened 2021-02-12
feature/add query support

FredrikOseberg:
Initial stab at adding project name filtering to the node SDK.

[x] Refactored url retrieval into urlGenerator to easily test
[x] Added projectName as optional options param
[x] Added tests for urlGenerator


src/urlGenerator.ts
```diff
@@ -0,0 +1,12 @@
+import { resolve } from 'url';
+
+class UrlGenerator {
```
ivarconr, line 3: it does not make sense to have a class for this util function. Maybe call it `url-utils`?

ivarconr:
I would like to see a test of the whole client, that it is adding the project query-param, using nock.

src/urlGenerator.ts
FredrikOseberg: Agree

ivarconr (approved):

PR #216, opened 2021-02-19
fix: use @vercel/fetch to better guard against network glitches

ivarconr:


ivarconr:
The disabled network test is not failing. A bit unsure on how to actually verify that the client still produces an error. (@vercel/fetch will do magic retries under the hood)

src/request.ts
```diff
@@ -69,7 +65,6 @@ export const post = ({ url, appName, timeout, instanceId, headers, json }: PostR
   fetch(url, {
```
ivarconr, line 65: we should probably also configure our own `retry` options.

SimenB:
At work we've set `retry: { retries: 3 }` and for the tests verifying behavior we make the call 4 times and expect the fourth to error out

ivarconr (commented):
Running it locally with a lot of traffic I got this error:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received null
    at validateString (internal/validators.js:120:11)
    at Url.parse (url.js:159:3)
    at urlParse (url.js:154:13)
    at fetchCachedDns (/home/ivarconr/code/unleash-client-node/node_modules/@vercel/fetch-cached-dns/index.js:62:30)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async /home/ivarconr/code/unleash-client-node/node_modules/@vercel/fetch-retry/index.js:52:23
    at async fetchRetry (/home/ivarconr/code/unleash-client-node/node_modules/@vercel/fetch-retry/index.js:48:14)
    at async vercelFetch (/home/ivarconr/code/unleash-client-node/node_modules/@vercel/fetch/index.js:74:11) {
  code: 'ERR_INVALID_ARG_TYPE',
  url: 'https://unleash.herokuapp.com/api/client/features',
  opts: {
    method: 'GET',
    timeout: 10000,
    headers: Headers { [Symbol(map)]: [Object: null prototype] },
    agent: HttpsAgent {
      _events: [Object: null prototype],
      _eventsCount: 4,
      _maxListeners: undefined,
      defaultPort: 443,
      protocol: 'https:',
      options: [Object],
      requests: {},
      sockets: {},
      freeSockets: [Object],
      keepAliveMsecs: 30000,
      keepAlive: true,
      maxSockets: Infinity,
      maxFreeSockets: 256,
      freeSocketKeepAliveTimeout: 15000,
      timeout: 10000,
      socketActiveTTL: null,
      createSocketCount: 2,
      createSocketCountLastCheck: 0,
      createSocketErrorCount: 0,
      createSocketErrorCountLastCheck: 0,
      closeSocketCount: 0,
      closeSocketCountLastCheck: 0,
      errorSocketCount: 0,
      errorSocketCountLastCheck: 0,
      requestCount: 58,
      requestCountLastCheck: 0,
      timeoutSocketCount: 0,
      timeoutSocketCountLastCheck: 0,
      maxCachedSessions: 100,
      _sessionCache: [Object],
      [Symbol(kCapture)]: false
    },
    redirect: 'manual',
    onRedirect: [Function]
  }
```

ivarconr:
It seems like `fetch-retry` passes custom retry options to fetch. This ends up destroying for nock, as it does not intercept the http calls when the method signature is incorrect. 


I have opened an issue for this:
https://github.com/vercel/fetch-retry/issues/40

ivarconr:
I have been testing this module and found quite a few bugs. 

1) 304 is not redirect. the included `@vercel/fetch-cached-dns` expects a "location" header set on the repsonse, and thus fails. https://github.com/vercel/fetch-cached-dns/issues/28
2) @vercel/fetch-retry adds internal (not compliant fetch) options to the fetch call if you override fetch per call. https://github.com/vercel/fetch-retry/issues/40

Conclusion: We can not use this wrapper to guard against network errors, as it will introduce new once. Especially 1) is a problem for us.

PR #217, opened 2021-02-21
feat: use make-fetch-happen instead of node-fetch

ivarconr:
In respect to some network issues discovered with node-fetch, where it does not automatically do reties for common network errors I have switched out node-fetch in favor of [make-fetch-happen](https://github.com/npm/make-fetch-happen)

I also played with the [got](https://www.npmjs.com/package/got) http client. The benefit of got is that it also support http2 and seems to be even more used in the wild. Switching to got will require us to to a larger rewrite and probably some more testing and verification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "homepage": "https://github.com/Unleash/unleash-client-node",
   "dependencies": {
     "ip": "^1.1.5",
+    "make-fetch-happen": "^8.0.14",
     "murmurhash3js": "^3.0.1",
     "node-fetch": "^2.6.1"
   },
@@ -43,9 +44,9 @@
   "devDependencies": {
     "@ava/babel": "^1.0.1",
     "@types/ip": "^1.1.0",
+    "@types/make-fetch-happen": "8.0.1",
     "@types/murmurhash3js": "^3.0.2",
     "@types/node": "^14.0.1",
-    "@types/node-fetch": "^2.5.8",
     "@typescript-eslint/eslint-plugin": "^4.14.2",
     "@unleash/client-specification": "^3.3.1",
     "ava": "^3.15.0",
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
   "dependencies": {
     "ip": "^1.1.5",
     "make-fetch-happen": "^8.0.14",
-    "murmurhash3js": "^3.0.1",
-    "node-fetch": "^2.6.1"
+    "murmurhash3js": "^3.0.1"
   },
   "engines": {
     "node": ">=10",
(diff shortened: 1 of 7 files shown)


chriswk (approved):
Seems good. Is the change of streams implementation (node core -> minipass) something we'd need to worry about?

FredrikOseberg (approved):
LGTM

package.json
```diff
@@ -29,6 +29,7 @@
   "homepage": "https://github.com/Unleash/unleash-client-node",
   "dependencies": {
     "ip": "^1.1.5",
+    "make-fetch-happen": "^8.0.14",
     "murmurhash3js": "^3.0.1",
     "node-fetch": "^2.6.1"
```
SimenB, line 34: is this still used?

test/unleash.network.retries.test.js
```diff
@@ -0,0 +1,39 @@
...
+  t.plan(1);
+
+  let calls = 0;
+  const server = createServer((req, res) => {
+    calls ++;
+    res.writeHead(408);
+    res.end();
+  });
+
+  server.listen(async () => {
```
SimenB, line 15:
```suggestion
  server.listen(() => {
```

test/unleash.network.retries.test.js
```diff
@@ -0,0 +1,39 @@
+import test from 'ava';
+import { createServer } from 'http';
+import { Unleash } from '../lib/unleash';
+
+test.cb('should retry on error', (t) => {
+  t.plan(1);
+
+  let calls = 0;
+  const server = createServer((req, res) => {
+    calls ++;
```
SimenB, line 10:
```suggestion
    calls++;
```

weird space, might just be me 😀

test/unleash.network.test.js
```diff
@@ -20,6 +18,7 @@ test.cb('should emit network errors', (t) => {
   });
 
   unleash.on('error', (e) => {
+    // console.log(e);
```
SimenB, line 21:
```suggestion
```

SimenB (approved):

package.json
ivarconr: no, will remove :+1:

src/request.ts
```diff
@@ -44,23 +44,24 @@ export const buildHeaders = (
...
+    head['If-None-Match'] = etag;
   }
   if (contentType) {
-    head.append('Content-Type', contentType);
+    head['Content-Type'] = contentType;
   }
   if (custom) {
-    Object.keys(custom).forEach((k) => head.append(k, custom[k]));
+    // eslint-disable-next-line no-return-assign
+    Object.keys(custom).forEach((k) => (head[k] = custom[k]));
```
SimenB, line 64:
```suggestion
    Object.assign(head, custom);
```

I think that'll work

PR #220, opened 2021-04-07
Normalized stickiness case to work properly with Gitlab

Darwex:
Gitlab send stickiness for strategies un uppercase, this small PR normalizes the name and makes it work with Unleash integrated in gitlab.


ivarconr:
Hi, thanks for reporting this. This sounds like a bug on the GitLab side of things, where they do not comply with the protocol. 

Just lower-casing it will work for the default and random case. But what about:
-  custom stickiness factors? 
- should all SDKs do this?

What about stickiness for variants? 

Could we reach out to GitLab and see if they can comply to the specifications instead?

glensc:
Did anyone manage to contact GitLab in this matter? perhaps link their issue or merge request?

Issue #226, opened 2021-05-20
add support for Repository.prototype.fetch to be able to receive request parameters


Albertobar94:
I had an issue with ca certificates since I am working with a custom ca from the company who I work for. Unfortunately, I have to modify the libraries code because I do not have an option include a custom certificate in the request to initialize unleash client... so basically in the post request using Repository.prototype.fetch the option to add ca would make it for this bug.

Albertobar94:
in the request.js file
var httpsAgent = new https.Agent({
    keepAlive: true,
    keepAliveMsecs: 30 * 1000,
    timeout: 10 * 1000,
    ca: <Value to receive>,
});

ivarconr:
We could potentially add support for custom SSL certificates. 

I wonder though, would it be possible to use the node option [NODE_EXTRA_CA_CERTS](https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file) to extend with your custom certificates as trusted by your application? I think this approach should work out of the box without changing the SDK.

Albertobar94:
I could not make it work setting the environment at the first launch with that since my Mac is using the Company's CN... so to make it easier I saw I could use a custom ca in that setting.. or even a 'rejectUnauthorized: false' and still have a secure environment since the connection making me problem is becauase of a SELF_SINGED_CERTIFICATE_IN_CHAIN and we use this library at work... maybe I can help out with a PR?

ivarconr:
> maybe I can help out with a PR

Yes please 👍

austinmehmet:
@ivarconr @Albertobar94 - Any update on this?

I tried providing the `NODE_EXTRA_CA_CERTS` env var and it doesn't seem to do anything with Unleash. I have to already set this var for connection to my company's internal APIs and it works for that. 

Being able to set `rejectUnauthorized` to false for localhost would be helpful

ivarconr:
I kind of lost track of this as I expected a PR for this. Would you be up for that @austinmehmet? If not I will try to schedule it for next week.

austinmehmet:
@ivarconr I can see if I can try and carve out some time to put up a PR.

ivarconr:
Thanks @austinmehmet

austinmehmet:
Whats weird is that I ran into a different SSL issue and tried doing 

`require('https').globalAgent.options.ca = require('fs').readFileSync('my/companies/certs.crt')`

[...]

PR #229, opened 2021-07-02
Add Support for Passing an HTTP Arguments Option for rejectUnauthorized

austinmehmet:
For #226  

Add ability to pass in an `HttpOptions` object with, as of now, only one field `rejectUnauthorized` to help with custom ca certs. The object was created if in the future other options need to be passed in such as an actual ca cert file, custom headers, proxy settings, etc.


austinmehmet:
@ivarconr - mind taking a look?

chriswk (approved):
This looks good to me, I'm missing a test though, though I'm inclined to ignore it since I'm not sure what you'd test.

Thank you

ivarconr (approved):
lgtm 👍🏼

PR #235, opened 2021-07-29
Changed the priority of the strategy block

Slauta:
**Problem, for example:**
It is necessary to change the logic of the `userWithId` strategy

In `client.ts`, the search for the current strategy is done using
```
return this.strategies.find ((strategy: Strategy): boolean => strategy.name === name);
```

**Solution**
Change the order of the merge
```
const strats = strategies.concat(defaultStrategies);
```

Thus, by changing the order of merging strategies, we are able to find our changed strategy with the name `userWithId`, if we embrace it in the configuration:
```
const client = initialize({
...
  strategies: [new UserWithIdStrategy()],
...
});
```

### DIFF
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -124,7 +124,7 @@ export class Unleash extends EventEmitter {
         httpOptions,
       });
 
-    const strats = defaultStrategies.concat(strategies);
+    const strats = strategies.concat(defaultStrategies);
 
     this.repository.on('ready', () => {
       this.client = new Client(this.repository, strats);


ivarconr:
Thanks, this makes sense.  Currently traveling, but will take a look next week if this is not picked up by the team by then.

ivarconr:
Do you mind also writing a test for this? This will document the functionality and ensure it is not switched back in a future release.

ivarconr (approved):

PR #237, opened 2021-08-05
Add support to provide namePrefix and tags for filtering

Mistic92:


src/url-utils.ts
```diff
@@ -1,10 +1,25 @@
...
+  if (namePrefix) {
+    params.append('namePrefix', namePrefix);
+  }
+  if (tags) {
+    tags.forEach((tag) => {
+      params.append('tag', tag);
+    });
+  }
+  return `${url}?${decodeURIComponent(params.toString())}`;
 };
```
ivarconr, line 23: Would you mind creating a unit test for this util as well?

src/url-utils.ts
```diff
@@ -1,10 +1,29 @@
...
 
-const getUrl = (base: string, projectName?: string): string => {
+const getUrl = (
+  base: string,
+  projectName?: string,
+  namePrefix?: string,
+  tags?: Array<string>,
+): string => {
+  const url = resolve(base, './client/features');
+  const params = new URLSearchParams(url);
```
ivarconr, line 10: you should not append `url` to the URLSearchParams. 

This ends up creating:
`http://some-url?http://someUrl=`

Simply replace this line with:
`const params = new URLSearchParams();`

test/url-utils.test.js
```diff
@@ -11,6 +11,31 @@ test('geturl should return url without project if projectname is not provided',
...
 });
 
+test('geturl should return url with namePrefix if namePrefix is provided', t => {
+  const result = getUrl('http://unleash-app.com', '', 'unleash');
+  t.true(result === 'http://unleash-app.com/client/features?namePrefix=unleash');
+});
+
+test('geturl should return url with namePrefix and project both are provided', t => {
+  const result = getUrl('http://unleash-app.com', 'myProject', 'unleash');
+  t.true(result === 'http://unleash-app.com/client/features?namePrefix=unleash&namePrefix=unleash');
```
ivarconr, line 21: Use the `t.is()` here will make it easier to spot the error. 

This line needs to change to:
`t.is(result, 'http://unleash-app.com/client/features?project=myProject&namePrefix=unleash');`

test/url-utils.test.js
```diff
@@ -11,6 +11,31 @@ test('geturl should return url without project if projectname is not provided',
...
+});
+
+test('geturl should return url with namePrefix, project and tags if all provided', t => {
+  const result = getUrl('http://unleash-app.com', 'myProject', 'unleash', ['tagName:tagValue']);
+  t.true(result === 'http://unleash-app.com/client/features?namePrefix=unleash&namePrefix=unleash&tag=tagName:tagValue');
```
ivarconr, line 26: error in the expected value.

test/url-utils.test.js
```diff
@@ -11,6 +11,31 @@ test('geturl should return url without project if projectname is not provided',
...
+});
+
+test('geturl should return url with tags if tags are provided', t => {
+  const result = getUrl('http://unleash-app.com', '', '', ['tagName:tagValue']);
+  t.true(result === 'http://unleash-app.com/client/features?tag=tagName:tagValue');
+});
+
+test('geturl should return url with two tags if two tags are provided', t => {
+  const result = getUrl('http://unleash-app.com', '', '', ['tagName:tagValue', 'tagName2:tagValue2']);
+  t.true(result === 'http://unleash-app.com/client/features?tag=tagName:tagValue&tag=tagName2:tagValue2');
```
ivarconr, line 36: I think this will work out of the box with express. In other examples we have been using empty square bracket for array bases query params like this:

`http://unleash-app.com/client/features?tag[]=tagName:tagValue&tag[]=tagName2:tagValue2`

src/url-utils.ts
Mistic92: Uh you are right. Probably left this after some testing. Will fix in next commit.

test/url-utils.test.js
Mistic92: What do you recommend to do with it?

ivarconr: I tried reading up on the standards and it seems to be quite unclear both `brackets` and `repeat` format seems to be supported by express (unleash-server).

I do however think it will be worth while to urlEncode the `tagName:tagValue` params

src/url-utils.ts
```diff
@@ -1,10 +1,29 @@
...
-    return resolve(base, `./client/features?project=${projectName}`);
+    params.append('project', projectName);
   }
-  return resolve(base, './client/features');
+  if (namePrefix) {
+    params.append('namePrefix', namePrefix);
+  }
+  if (tags) {
+    tags.forEach((tag) => {
+      params.append('tag', tag);
```
ivarconr, line 19: I think we should do a `encodeURIComponent` like this:

`params.append('tag', encodeURIComponent(tag));`

ivarconr (commented):
Nice to get this in to the SDK 👍🏼 
Only fix the encoding of the tag values and I think we are good to go!

ivarconr (approved):
I removed the extra decodeURIComponent from `uri-utils.ts` that should not have been there. All good now!

PR #283, opened 2021-12-27
feat: add bootstrap

ykhedher:
This PR brings two new features to increase the resilience options for the Unleash Node SDK. 

**1. Bootstrap feature ([docs](https://www.getunleash.io/blog/bootstrap-unleash-java-sdk-to-improve-the-resilience))**
 - Allows you to bootstrap the SDK with feature toggle configuration coming from `data array`, `a file` or `a url`. This allows you to bootstrap the initial config from anywhere (potentially a locally hosted Unleash Proxy). 

**2. Custom Storage Implementation**
- Allows you to swap the file-based backup mechanism with something else, could be in-memory, redis, postgres, s3 or whatever makes sense for your application. This allows the SDK to persist states also in containerized environments. 

### TODO

- [x] Write documentation for boostrap feature
- [x] Write documentation for custom storage feature
- [x] Write unit test validating the following rules:
  - [x]  Backup should never override actual data from the Unleash API (live data)
  - [x]  Boostrap should never override actual data from the Unleash API (live data)
  - [x]  Boostrap may override backup, based on config.


ykhedher:
[Add boostrap to node.js SDK](https://trello.com/c/whGI2KJ6/665-add-boostrap-to-nodejs-sdk)

src/helpers.ts
```diff
@@ -25,3 +26,25 @@ export function resolveContextValue(context: Context, field: string) {
...
+  let info;
+  try {
+    info = userInfo();
+  } catch (e) {
+    // unable to read info;
+  }
+
+  const prefix = info
+    ? info.username
+    : `generated-${Math.round(Math.random() * 1000000)}-${process.pid}`;
```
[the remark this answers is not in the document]
ivarconr: @sonatype-lift ignore. This is not meant as a secure cryptographically string.

src/feature.ts
```diff
@@ -10,3 +10,9 @@ export interface FeatureInterface {
   strategies: StrategyTransportInterface[];
   variants: VariantDefinition[];
 }
+
+export interface ClientFeaturesResponse {
```
ivarconr, line 14: Any help for a better name for this appreciated! It is essentially the interface for the data returned by the Unleash Client SDK API.

src/metrics.ts
```diff
@@ -87,11 +88,6 @@ export default class Metrics extends EventEmitter {
     this.timeout = timeout;
     this.resetBucket();
     this.httpOptions = httpOptions;
-
-    if (typeof this.metricsInterval === 'number' && this.metricsInterval > 0) {
```
ivarconr, line 91: We should not start tasks in the constructor of a class, this reduces flexibility and may cause us to miss events because the using code is not able to register event listeners before the event actually happens.

src/strategy/remote-addresss-strategy.ts
```diff
@@ -12,20 +12,18 @@ export default class RemoteAddressStrategy extends Strategy {
...
-          } catch (err) {
-            return false;
-          }
-        }
-        return false;
-      },
-    );
+      }
+      return false;
+    });
```
ivarconr, line 27: this file was automatically reformatted by the lint-setup.

src/unleash.ts
```diff
@@ -7,13 +7,16 @@ import { CustomHeaders, CustomHeadersFunction } from './headers';
...
-import { FallbackFunction, createFallbackFunction } from './helpers';
+import { FallbackFunction, createFallbackFunction, generateInstanceId } from './helpers';
 import { HttpOptions } from './http-options';
 import { TagFilter } from './tags';
+import { BootstrapOptions, resolveBootstrapProvider } from './repository/bootstrap-provider';
+import { FileStorageProvider, StorageProvider } from './repository/storage-provider';
+import { UnleashEvents } from './events';
 
-export { Strategy };
+export { Strategy, UnleashEvents };
```
ivarconr, line 19: TODO: We need to export these from index.ts.

examples/bootstrap.js
```diff
@@ -0,0 +1,75 @@
...
+   url: 'http://localhost:3000/proxy/client/features',
+   urlHeaders: {
+     Authorization: 'bootstrap',
+   }
+  },
+});
+
+client.on('error', () => console.log("\x1b[31m", 'Unable to fetch feature toggles', "\x1b[0m"));
+client.on('warn', console.log);
+// client.on('changed', () => console.log('changed'));
```
sighphyre, line 54: Should we trim out the commented code? Unless it's adding something useful to understanding what this is doing, I vote we remove it

examples/custom_storage.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable */
+const { get } = require('http');
+const { initialize } = require('../lib');
+
+const { createClient } = require('redis');
+
+class MyRedisStore {
```
sighphyre, line 7: Cool!

src/helpers.ts
```diff
@@ -25,3 +26,25 @@ export function resolveContextValue(context: Context, field: string) {
...
+}
+
+
+export function generateInstanceId(instanceId?: string): string {
+  if(instanceId) {
+    return instanceId;
+  } 
+  let info;
+  try {
+    info = userInfo();
```
sighphyre, line 41: Can we console.warn this or something? I've spent too much of my life hunting down exceptions that died in an empty catch block.

sighphyre (approved):
Very cool!

examples/bootstrap.js
ivarconr: Yes, indeed!

src/helpers.ts
ivarconr: fair point. Its not super important feature, but it can be quite nice way to identify the actual connected instance.

ivarconr: after thinking about this I would rather not issue an warning here. This will often just work in a local environment, but still very useful as you will end up seeing the username for that computer in the connected instance id.

PR #289, opened 2022-01-20
feat: add support for new constraint operators

ivarconr:


src/strategy/strategy.ts
```diff
@@ -28,11 +31,33 @@ export class Strategy {
...
+      }
+      case Operator.STARTS_WITH: {
+        return constraint.values.some((val) => contextValue.startsWith(val.trim()));
+      }
+      case Operator.CONTAINS: {
+        return constraint.values.some((val) => contextValue.includes(val.trim()));
+      }
+      default: 
+        return false;
+    }
```
ivarconr, line 59: Instead of a switch on the Operator-enum I am tempted to instead create a small handler function per operator and just execute that function. 

Example for contains:

```
const handleContains = (values: string[], contextValue) => values.some((val) => contextValue.includes(val.trim()))
```

ivarconr: Also, we should probably pre-map the values by trimming them, making the handler work on pure data. 

```ts
values.map((v) => v.trim())
````

File: src/helpers.ts [hunk 2]
```diff
@@ -16,12 +17,34 @@ export function createFallbackFunction(
...
+  let info;
+  try {
+    info = userInfo();
+  } catch (e) {
+    // unable to read info;
+  }
+
+  const prefix = info
+    ? info.username
+    : `generated-${Math.round(Math.random() * 1000000)}-${process.pid}`;
```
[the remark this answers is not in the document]
ivarconr: @sonatype-lift ignore

chriswk (approved):
This looks really good. Adds a lot of value to constraints :)

src/strategy/strategy.ts
```diff
@@ -10,14 +11,139 @@ export interface StrategyTransportInterface {
...
+  } if(operator === Operator.SEMVER_GT) {
+    return semverGt(contextValue, value); 
+  } 
+    return false;
+}
+
+const DateOperator = (constraint: Constraint, context: Context) => {
+  const { operator } = constraint;
+  const value = new Date(constraint.value as string);
+  const currentTime = context.currentTime ? new Date(context.currentTime) : new Date();
```
ivarconr, line 97: we might want to move this to the creation of the context. 

+ The benefit will be that the date and time will not change between multiple constriants within the same feature toggle. 
- The downside would be that we would have to create the currentTime date instance on the context, even when a use is not using any constraints that uses date operators.

PR #307, opened 2022-03-17
Add forceGetVariant method

sighphyre:
Adds a forceGetVariant method which will not check the state of the parent toggle when called. This is to fix an issue in the ProxySDK, which currently resolves variant enabled state twice, which causes the % of a given variant to incorrectly resolve.

This isn't covered with unit tests, if someone can suggest a viable unit/integration test that would be great, given the statistically variant nature of the issue, it has been manually tested by manually linking the proxy sdk with the following Python script and checking visually that the distributions are within tolerance:

```
import requests

max_requests = 100

no_toggles = 0
disabled_variant = 0
var1 = 0
var2 = 0

for i in range(0, max_requests):
    r = requests.get('http://localhost:3000/proxy', headers={"Authorization":"proxy-secret"})
    response = r.json()
    toggles = response["toggles"]

    if toggles:
        toggle = toggles[0]
        variant = toggle.get('variant')
        if variant:
            name = variant.get('name')
            if name == "disabled":
                disabled_variant += 1
            elif name == "Var1":
                var1 += 1
            elif name == "Var2":
                var2 += 1
            else:
                raise Exception("Unexpected variant")
    else:
        no_toggles += 1
print("No toggles:", no_toggles, no_toggles/max_requests)
print("Disabled variant:", disabled_variant, disabled_variant/max_requests)
print("Var1:", var1, var1/max_requests)
print("Var2:", var2, var2/max_requests)
print("---")
print("Total:", (var1 + var2 + no_toggles + disabled_variant)/max_requests)
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -95,6 +95,22 @@ export default class UnleashClient extends EventEmitter {
   }
 
   getVariant(name: string, context: Context, fallbackVariant?: Variant): Variant {
+    return this.resolveVariant(name, context, true, fallbackVariant);
+  }
+
+  // This function is intended to close an issue in the proxy where feature enabled
+  // state gets checked twice when resolving a variant with random stickiness and
+  // gradual rollout. This is not intended for general use, prefer getVariant instead
+  forceGetVariant(name: string, context: Context, fallbackVariant?: Variant): Variant {
+    return this.resolveVariant(name, context, false, fallbackVariant);
+  }
+
+  private resolveVariant(
+    name: string,
+    context: Context,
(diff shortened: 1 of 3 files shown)


thomasheartman (approved):
Code looks good to me! I agree that it'd be great to get this tested somehow, but I'm not sure exactly how. 

Can you do the same thing as in the python script, just in a test? 🤔  The endpoint should behave the same regardless, right?

ivarconr (approved):
Nice :D

PR #318, opened 2022-04-19
Update Documentation - Clearer client alternative usage and highlight the require use of error handling

cristobal:
## Background
PR based on the issue here https://github.com/Unleash/unleash-client-node/issues/310

## What's Done
* Move the alternative usage section further up in the documentation as a subsection of the Initialize unleash-client section.
* Highlight in a note that when using unleash directly, that its required to add an event listener and if it is not done the application may either not start up or crash if it can't reach the unleash server.
* Remove defunct note about `require.cache`

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,27 @@ unleash.on('synchronized', () => {
   const variant = unleash.getVariant('app.ToggleY');
 });
 ```
+Be aware that the `initialize` function will configure a _global_ Unleash instance. If you call this method multiple times the global instance will be changed. 
 
-Be aware that the `initialize` function will configure a _global_ Unleash instance. If you call this
-method multiple times the global instance will be changed. If you prefer to handle the instance
-yourself you should [construct your own Unleash instance](#alternative-usage).
+#### <u>Constructing the Unleash client directly</u>
+If you prefer to construct the Unleash instance yourself rather than using the global `initialize` method you can do so e.g.:
+```js
+const { Unleash } = require('unleash-client');
+
+const unleash = new Unleash({
+  appName: 'my-app-name',
+  url: 'http://unleash.herokuapp.com',
+  customHeaders: {
+    Authorization: 'API token',
+  },
+});
+
+unleash.on('ready', console.log.bind(console, 'ready'));
+
+// required error handling when using unleash directly  
+unleash.on('error', console.error);
+```
+**_IMPORTANT:_** When using unleash directly, error handling is required by adding an event listener for the error event. Otherwise your application may either not start up, or crash if its running when the application can't reach the unleash server.
 
 #### Block until Unleash SDK has synchronized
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,10 +46,27 @@ unleash.on('synchronized', () => {
   const variant = unleash.getVariant('app.ToggleY');
 });
 ```
+Be aware that the `initialize` function will configure a _global_ Unleash instance. If you call this method multiple times the global instance will be changed. 
 
-Be aware that the `initialize` function will configure a _global_ Unleash instance. If you call this
-method multiple times the global instance will be changed. If you prefer to handle the instance
-yourself you should [construct your own Unleash instance](#alternative-usage).
+#### <u>Constructing the Unleash client directly</u>
```
thomasheartman, line 51: Is the `u` tag intended to underline the heading? It doesn't seem to work on GitHub (at least not in the preview), and I think we should leave the header styling up to the markdown renderer anyway 💁🏼

```suggestion
#### Constructing the Unleash client directly
```

README.md
```diff
@@ -46,10 +46,27 @@ unleash.on('synchronized', () => {
...
+If you prefer to construct the Unleash instance yourself rather than using the global `initialize` method you can do so e.g.:
```
thomasheartman, line 52: Nice! When would you want to do this / what problems does it solve? I think giving the users an example of when they'd want to do this would be useful 🤓

Also, a slight rewording suggestion:
```suggestion
You can also construct the Unleash instance yourself instead of via the `initialize` method. 
This lets you [...] // <- why is this useful?
```

README.md
```diff
@@ -46,10 +46,27 @@ unleash.on('synchronized', () => {
...
+    Authorization: 'API token',
+  },
+});
+
+unleash.on('ready', console.log.bind(console, 'ready'));
+
+// required error handling when using unleash directly  
+unleash.on('error', console.error);
+```
+**_IMPORTANT:_** When using unleash directly, error handling is required by adding an event listener for the error event. Otherwise your application may either not start up, or crash if its running when the application can't reach the unleash server.
```
thomasheartman, line 69: This is great! But I think it might make sense to move it to before the code sample, so that it's:
1. introductory paragraph
2. error handling
3. code sample

What do you think? ☺️ 

Also a minor rewording suggestion:

```suggestion
**Important:** When Using unleash directly, you **must** handle errors yourself. This can be done by attaching an event listener to the `error` event, as shown in the example below. If you don't do this, your application may crash either on startup or at runtime if it can't reach the Unleash server or if it encounters other errors.
```

thomasheartman (changes requested):
Super nice! 🙌🏼 Thanks for this. I've got a few suggestions added inline, but overall I think this is great work 😄

README.md
cristobal: Yeah the idea was to stylize it so it looked like a proper subsection of section `2.1`, but not that important anyways 👍🏽 
My markdown renderer locally did render the underline properly so  i thought it would work on Github too 🤷🏽

README.md
cristobal: > This is great! But I think it might make sense to move it to before the code sample, so that it's:
> 
> 1. introductory paragraph
> 2. error handling
> 3. code sample
> 
> What do you think? ☺️
I agree will change the order of things 👍🏽

cristobal: > 
> Also a minor rewording suggestion:
Thanks for the rewording suggestion, gave this text a couple of rounds but thought that it lacked something that's what PR's for 🙂  it reads much better now 👍🏽

ivarconr (approved):

PR #320, opened 2022-04-28
fix: Use string[] instead of String[] for Override values

thomasheartman:
This PR changes the use of `String[]` to the use of `string[]` in the variant type definition. It is assumed that this is just done in error and doesn't actually carry any meaning.

This seems to be the only case where we're using uppercase Strings.
It's causing issues with OpenAPI typing, which says that String is not
string and that it can't compile with this issue.

### DIFF
--- a/src/variant.ts
+++ b/src/variant.ts
@@ -10,7 +10,7 @@ enum PayloadType {
 
 interface Override {
   contextName: string;
-  values: String[];
+  values: string[];
 }
 
 export interface Payload {


chriswk (approved):
Yes, good spot, definitely string[] we want to use :)

thomasheartman:
@chriswk Wait, before I merge: is this a breaking change? The API will change. Is there any cases where this might break something for the end user?

thomasheartman:
@chriswk Wait, before I merge: is this a breaking change? The API will change. Is there any cases where this might break something for the end user?

thomasheartman:
Then again, if it _is_ a bug, then maybe it's fine 💁🏼

thomasheartman:
Then again, if it _is_ a bug, then maybe it's fine 💁🏼

PR #328, opened 2022-05-03
Type is missing from FeatureInterface

mathjoh:
The FeatureInterface was missing the type member, despite the data containing it when using the getFeatureDefinitions function. 

I'm making a feature where an app gets all feature toggles of type experiment with a certain naming scheme. And it would be nice to use the type member without creating an extension interface in my code.


chriswk (approved):
Thank you for this. I do believe you're right, we're already fetching this in queries, so exposing it in the client feels like the right thing to do.

PR #329, opened 2022-05-09
feat: Add support for handling global segments

sighphyre:
This implements support for handling global segments over inline segments, related to this PR on the Unleash server: https://github.com/Unleash/unleash/pull/1577 (which needs to be merged before this can do anything useful). 

This has been reworked slightly to improve concerns around memory. This now precomputes a Map<number, Constraint> on resolving toggles and uses Generator objects to lazily resolve the relevant constraint values. This means no more costly array spreading or memory copying and this now retains the time complexity characteristics of the previous solution while having significantly reduced memory usage.

(If you're not familiar with generators the MDN docs are quite useful: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator)


sighphyre:
[Segmentation SDK PoC - Implement solution in one SDK as a reference implementation](https://trello.com/c/JsQAa5rd/1104-segmentation-sdk-poc-implement-solution-in-one-sdk-as-a-reference-implementation)

examples/simple_usage.js
```diff
@@ -1,12 +1,13 @@
...
-const unleashContext = {userId: '1232'};
+const url = 'http://localhost:4242/api/';
+const apiToken = '*:development.5ec41f4a67313faa2499a33b56a4d8aa1547c7ad1ef2af8f517f0fa5';
+const toggleName = 'AWholeNewToggle';
+const unleashContext = {
+};
 
 const client = initialize({
-  appName: 'my-application',
+  appName: 'BOB',
```
olav, line 10: :eyes:

src/repository/index.ts
```diff
@@ -77,6 +78,8 @@ export default class Repository extends EventEmitter implements EventEmitter {
 
   private data: FeatureToggleData = {};
 
+  private knownSegments?: Segment[];
```
olav, line 81: Could this be `segments`? Not sure what `known` implies.

src/repository/index.ts
```diff
@@ -288,11 +295,38 @@ Message: ${err.message}`);
     this.removeAllListeners();
   }
 
+  resolveSegment(segmentId: number): Segment | undefined {
```
olav, line 298: Maybe `findSegmentById`.

src/repository/index.ts
```diff
@@ -288,11 +295,38 @@ Message: ${err.message}`);
...
+    return this.knownSegments?.find((segment) => segment.id === segmentId);
+  }
+
+  unpackSegments(feature: FeatureInterface): FeatureInterface {
```
olav, line 302: Maybe `inlineSegmentConstraints`?

src/strategy/strategy.ts
```diff
@@ -17,6 +18,15 @@ export interface Constraint {
   caseInsensitive?: boolean;
 }
 
+export interface Segment {
+  id: number,
+  name: string,
+  description?: string,
+  constraints: Constraint[],
+  createdBy: string,
+  createdAt: Date
```
olav, line 27: Should `createdAt` be a string or do we automatically parse dates somewhere?

src/repository/index.ts
```diff
@@ -288,11 +295,38 @@ Message: ${err.message}`);
...
+            const foundSegment = this.resolveSegment(segmentId);
+            const constraints = foundSegment?.constraints || [];
+            return constraints;
+          })
+          ?.reduce((prev, next) => prev.concat(next), []) || [];
+      const safeConstraints = strategy?.constraints || [];
+      const safeSegmentConstraints = segmentConstraints || [];
+      return {
+        ...strategy,
+        constraints: [...safeConstraints, ...safeSegmentConstraints],
```
olav, line 319: This still seems potentially heavy when there are lots of segments and toggles, but not sure if too heavy.

examples/simple_usage.js
sighphyre: Oh my word. Thank you. Rolled that back

src/repository/index.ts
sighphyre: It could be segments! So the known is there because this is not all the segments but only the ones that are actively used. But the node sdk doesn't actually really know that or even really care about that. So... good point, I'll change this to segments

src/repository/index.ts
sighphyre: Yep, that's better, thanks!

src/repository/index.ts
sighphyre: Yeah, I thought that too so I ran some local benchmarks. Spreading two arrays like this with a million items in each takes around 30 milliseconds (on my machine). That's probably a JSON packet measured in tens or hundreds of MB. I want to say that's probably fast enough for an excessively large packet, what do you think?

olav: True, that seems fine. I was thinking about memory since each segment gets copied into all its features. Maybe for later.

sighphyre: That, sir, is an excellent point! And I think it's less of a "this might be heavy" and more of a "at some point this is going to be too heavy". The questions become, at what point is it too heavy, is that point within a range we want to achieve and will anything else fall over before this does. I don't have answers to those things right now, so the tradeoff I've made here is to accept that some memory balooning will happen during resolution here in exchange for much less invasive changes to the SDK. Let me setup a few benchmarks here and poke over the next few days and let some others take a look as well

src/strategy/strategy.ts
sighphyre: Awesome, it's a string now!

src/repository/index.ts
ivarconr: I still think we need to have some limits for segments. If we assume a segment will never contain more than, say "a few thousand items", would this be a concern? Assuming number of feature toggles are less than 1000?

ivarconr:
I remember generator objects from my Ruby on Rails days! 🗡️

src/client.ts
```diff
@@ -85,15 +86,41 @@ export default class UnleashClient extends EventEmitter {
           this.warnOnce(strategySelector.name, feature.name, feature.strategies);
           return false;
         }
-        return strategy.isEnabledWithConstraints(
-          strategySelector.parameters,
-          context,
-          strategySelector.constraints,
-        );
+        const constraints = this.yieldConstraintsFor(strategySelector);
```
ivarconr, line 89: this is quite elegant !

ivarconr (approved):
I like the approach. Looks nice and clean.

src/client.ts
sighphyre: Thank you! I'm a fan as well, generators make this really easy to do

PR #330, opened 2022-05-10
docs: fix example usage in readme

Tymek:
Fix typo in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ unleash.once('changed', () => {
     console.log(`Demo is enabled: ${unleash.isEnabled('Demo')}`);
 });
 
-unleash.on('count', (name, enabled) => console.log(`isEnabled(${name}`));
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -231,7 +231,7 @@ unleash.once('changed', () => {
     console.log(`Demo is enabled: ${unleash.isEnabled('Demo')}`);
 });
 
-unleash.on('count', (name, enabled) => console.log(`isEnabled(${name}`)
+unleash.on('count', (name, enabled) => console.log(`isEnabled(${name}`));
```
thomasheartman, line 234: Nice catch! But I think there's another closing parenthesis missing 🤓  Like this, maybe?
```suggestion
unleash.on('count', (name, enabled) => console.log(`isEnabled(${name})`));
```

thomasheartman (approved):
Good catch! But I think there as a little bit more missing: see the inline comment

PR #332, opened 2022-06-02
Expose fetch/agent option via HttpOptions

yflander:
## About the changes

Added the ability to provide a custom HTTP agent. In our case, we would like to use an agent with DNS caching. More details on general Node.js issue with DNS: https://httptoolkit.tech/blog/configuring-nodejs-dns/

Closes # .

### DIFF
--- a/src/http-options.ts
+++ b/src/http-options.ts
@@ -1,3 +1,7 @@
+import type { URL } from 'url';
+import type { Agent } from 'http';
+
 export interface HttpOptions {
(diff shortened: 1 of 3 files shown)


jameslnewell:
Nice! Can we submit a PR upstream too 🙏

jameslnewell (approved):

olav (approved):
@yflander Thanks for the PR, looks good to me.

PR #333, opened 2022-06-02
fix: force semver operators to resolve false against non strict semver

sighphyre:
This patches the semver constraints to resolve to false when the passed in context is not a strict semver, in order to bring this sdk up to speed with the latest version of the client specification tests


thomasheartman (approved):
Nice change! 🥳 If that's needed to pass the spec, then it's good for sure 💯  Will this require a new major version or does it get away with it?

ivarconr (approved):

sighphyre:
@thomasheartman I think it gets away with it in this case, we won't be changing any API comparability here, just making it do what it ought to be doing. I'll be including this in a minor patch to also release global segments but if we need we can always backport this to older versions

PR #339, opened 2022-06-15
Use fallback value if not initialised

### LINKED ISSUE #338: FallbackValue not used if instance not initialised
### Describe the bug

In https://github.com/Unleash/unleash-client-node/pull/13 support for `fallbackValue` was introduced. But the fallback is only returned if there is an [Unleash instance](https://github.com/Unleash/unleash-client-node/blob/main/src/index.ts#L32). If `initialize` (or `startUnleashed`) is not called, the fallback value is ignored and returns always `false`.

I would use this for local development. We don’t want to initialise Unleash so we are independent from the current flags on the server. Instead the defined default should be used when running locally or in a local docker container.

* call `isEnabled` without calling `initialize` before.

**Example:**
```typescript
if (config.instanceId) {
    await startUnleash(config);
}
const featureIsEnabled = isEnabled("my-feature", {}, true);
```

If `config.instanceId` is not set, `featureIsEnabled` will be `false` ignoring the fallback `true`

### Expected behavior

Should use the fallbackValue, if instance is not initialised.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

unleash-client@3.14.1

### Subscription type

Open source

### Hosting type

Self-hosted

_No response_

LukasFritzeDev:
## About the changes
* The first commit fixes prettier issues on the current main branch. If wanted, I could remove this commit from my PR.
* The second commit sets the instance global to `undefined` on `destroy()`. This is required to make the tests in the third command work. I don’t think this is braking, since it would be the expected behaviour anyway.
* The third commit introduces the actual fix and sets the `fallbackValue` if the instance was not initialised before. (Tests including). For details see #338. 

Closes #338 

## Discussion points

* Should prettier changes to related to fix be included in this PR?

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -31,11 +31,10 @@ export function safeName(str: string = '') {
   return str.replace(/\//g, '_');
 }
 
-
 export function generateInstanceId(instanceId?: string): string {
-  if(instanceId) {
+  if (instanceId) {
     return instanceId;
-  } 
+  }
   let info;
   try {
     info = userInfo();
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,7 @@ import { Context } from './context';
 import { TagFilter } from './tags';
 import { UnleashEvents } from './events';
 import { ClientFeaturesResponse } from './feature';
-import InMemStorageProvider from './repository/storage-provider-in-mem'
+import InMemStorageProvider from './repository/storage-provider-in-mem';
 
 // exports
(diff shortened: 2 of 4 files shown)


ivarconr:
Thanks for providing this. Most of the changes makes sense. We just need some time to think/discuss the impact of the third change. It seems reasonable on first impression.

ivarconr (approved):

PR #340, opened 2022-06-27
fix: NOT_IN operator for missing contextField should return true.

ivarconr:
We saw some inconsistent handling of missing context fields on in the SDK together with `IN` and `NOT_IN` operator. 

E.g. When the user has a `NOT_IN` and does not provide any context field the SDK should return true, while the node SDK is currently always returning false in that case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "@types/node": "14.18.16",
     "@types/semver": "7.3.9",
     "@typescript-eslint/eslint-plugin": "5.10.0",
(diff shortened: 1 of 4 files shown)


ivarconr:
[Medidata: Add spec for missing custom context field on context, used in a constraint.](https://trello.com/c/FtEwCVhp/1323-medidata-add-spec-for-missing-custom-context-field-on-context-used-in-a-constraint)

sighphyre (approved):
LGTM

olav (approved):

PR #343, opened 2022-06-27
fix: string operator needs to guard against non-string values

ivarconr:
If the user passes a number instead of a string on a context field where we use the string operator the node SDK will currently crash. 

This PR adds a check that the context field is actually a string.


sighphyre (approved):
LGTM

PR #353, opened 2022-07-13
Missing http agent in bootstrap

CaptainGlac1er:
## About the changes

Added the option to add in http agent to bootstrap if wanted?

Closes # .


sighphyre (changes requested):
I think this looks great! Happy to merge this but would you mind adding a test to validate this works correctly?

PR #368, opened 2022-09-26
fix: Only initailize the SDK once.

ivarconr:
## About the changes
Before this fix all calls to `initialize` would create a new SDK instance, all maintaining the connection towards the Unleash API. We have observed this causing accidental DDosing of the Unleash API.

This fix will make sure that the SDK is only initialized once (aka singleton) when using the `initialize` method.

## Discussion points
Unsure if we should `warn` when the Unleash SDK has been **initialized more than 10 times**.


src/unleash.ts
```diff
@@ -185,6 +197,16 @@ export class Unleash extends EventEmitter {
     if (!disableAutoStart) {
       process.nextTick(async () => this.start());
     }
+    Unleash.instance = this;
+  }
+
+  static getInstance(config: UnleashConfig) {
```
ivarconr, line 203: Introducing the singleton pattern here.

src/unleash.ts
```diff
@@ -185,6 +197,16 @@ export class Unleash extends EventEmitter {
...
+    if(Unleash.instance) {
+      return Unleash.instance;
+    }
```
thomasheartman, line 206: If you call `getInstance` with a different config, does it make sense to update the config for the instance? Or is that just confusing? It feels a little misleading that you can pass in the config as an argument, but that it does nothing. That said, I'm not sure there is a better way? Could we make it parameterless somehow and have you pass config in other ways? 🤔

thomasheartman (changes requested):
Nice work and good call 🙌🏼  I've got a couple questions regarding how we should use the singleton inline plus some minor suggestions on wordings.

src/unleash.ts
ivarconr: Good question. I think that you are spot on the `getInstance()` method in the singleton pattern usually does not take arguments (options). I am a bit unsure how/where to communicate the options in a clear and concise way.

README.md
```diff
@@ -46,7 +46,7 @@ unleash.on('synchronized', () => {
   const variant = unleash.getVariant('app.ToggleY');
 });
 ```
-Be aware that the `initialize` function will configure a _global_ Unleash instance. If you call this method multiple times only the first call will setup a global Unleash instance. You may call `destroy()` to remove the globally configured instance.
+The `initialize` function will configure and creates a _global_ Unleash instance. When a global instance exists, calling this method has no effect. Call the `destroy` function to remove the globally configured instance.
```
thomasheartman, line 49: Nice! Thanks for accepting the suggestion. I think I made a mistake when suggesting it, so I've added a little fix here 😅 

```suggestion
The `initialize` function will configure and create a _global_ Unleash instance. When a global instance exists, calling this method has no effect. Call the `destroy` function to remove the globally configured instance.
```

src/unleash.ts
thomasheartman: Yeah, good point. If you remove the need completely and use default values, then I imagine almost everyone will just _not_ configure the instance at all.

src/unleash.ts
```diff
@@ -87,6 +91,14 @@ export class Unleash extends EventEmitter {
   }: UnleashConfig) {
     super();
 
+    Unleash.instanceCount++;
+
+    if(Unleash.instanceCount > 10) {
+      console.warn('The unleash SDK has been initialized more than 10 times')
+    }
+
```
thomasheartman, line 99: Regarding whether we should warn or not. I'm into the idea, but maybe make it configurable? For instance, add a `disableInstanceCountWarning` variable that can turn it off or something? 

Also, should it be `console.warn` or a configurable logger? I don't know how you usually do that in Node, but I'd expect to be able to redirect warnings to where I'd want them, I think. Also, if it's a configurable logger (or just a `logWarning: (warning: string) => void` function, then you could pass `() => {}` as an argument and disable the logging that way 🤷🏼

ivarconr: Good point. We currently do not support a log abstraction, and I think it would be more work to get that in than I am willing to do for this change. We do have the event emitter, but the purpose here is to want users of the SDK that something is off. 

A suitable middle-ground would be to add the option to disable the warning entirely, and log it as an error to the event emitter, which default will console.error it, and the user can override that by registering an on("error") handler.

thomasheartman: > I think it would be more work to get that in than I am willing to do for this change

Aye, I agree with this.

> A suitable middle-ground would be to add the option to disable the warning entirely, and log it as an error to the event emitter, which default will console.error it, and the user can override that by registering an on("error") handler.

Just to make sure I understand this: does it only log to the event emitter if you disable the warning, or does it log to the event emitter regardless? And do we have support "warning" events? This isn't an error, so logging it as such feels a little icky to me.

src/unleash.ts
sighphyre: What about something like:

```
static buildInstance(options?: UnleashConfig){
  if(!options){
     Unleash.singletonConfig = makeDefaultOptions();
  }else {
     Unleash.singletonConfig = options
  }
}

static getInstance() {
  if(!Unleash.singletonConfig){
    throw new Error("You gotta call buildInstance first")
  }
  //setup and/or return instance
}
```

ivarconr: you always makes me learn something new @sighphyre, thanks!

thomasheartman:
It's been a little while: are we still working on this PR or should we close it or otherwise?

src/unleash.ts
ivarconr: > Just to make sure I understand this: does it only log to the event emitter if you disable the warning

No, the plan is then to skip the error entirely,

examples/simple_usage.js
```diff
@@ -14,6 +14,19 @@ const client = initialize({
   }
 });
 
+const client2 = initialize({
```
ivarconr, line 17: todo: remove this again!

src/unleash.ts
```diff
@@ -84,9 +93,21 @@ export class Unleash extends EventEmitter {
...
     storageProvider,
     disableAutoStart = false,
+    skipInstanceCountWarning = false,
   }: UnleashConfig) {
     super();
 
+    Unleash.instanceCount++;
+
+    this.on(UnleashEvents.Error, (err) => {
+      console.error(err?.message);
```
ivarconr, line 103: we need to check if there are other event handlers registered and only console.error if no others have registered.

src/unleash.ts
```diff
@@ -84,9 +93,21 @@ export class Unleash extends EventEmitter {
...
+    });
+
+    if(!skipInstanceCountWarning && Unleash.instanceCount > 10) {
+      const error = new Error('The unleash SDK has been initialized more than 10 times');
```
ivarconr, line 107: need to wrap this in a process.nextTick() to ensure using code have a chance to register an event listener before it triggers.

src/unleash.ts
```diff
@@ -50,6 +53,12 @@ export interface StaticContext {
 }
 
 export class Unleash extends EventEmitter {
+  private static configSignature?: string;
+
+  private static instance?: Unleash;
+
+  private static instanceCount: 0;
```
ivarconr, line 60: Oh, this is a very nice bug, I accidentally set the type to "0" which is definetley not what I tried to do here!

src/unleash.ts
```diff
@@ -189,6 +217,26 @@ export class Unleash extends EventEmitter {
...
+  /**
+   * Will only give you an instance the first time you call the method, 
+   * and then return the same instance. 
+   * @param config The Unleash Config. 
+   * @returns the Unleash instance
+   */
+  static getInstance(config: UnleashConfig) {
+    const configSignature = generateHashOfObject(config);
+    if(Unleash.instance) {
+      if(configSignature !== Unleash.configSignature) {
```
sighphyre, line 229: This is pretty elegant, I like it

PR #376, opened 2022-10-19
Emit "unchanged" event

johnjcsmith:
## About the changes
`Unleash` was not actually re-emitting the "unchanged" event from the toggle repository as expected based on the readme.
Add an event handler on `repository` to forward this event to the consumers.


thomasheartman:
Hey, @johnjcsmith 👋🏼 Thanks for the PR! It looks like it's doing the right thing to me. Let me just check with someone who's more familiar with how the events work in this SDK to make sure it's all good 😄

chriswk (approved):
LGTM

PR #392, opened 2022-12-22
fix: handle socket hang up gracefully.

ivarconr:
If the SDK gets a socket hang up we will not get
a built in retry from make-fetch happen.

This fix will issue a new fetch in refreshInterval / 2 in order to handle this with a graceful feedback.


chriswk (approved):
Nice!

PR #412, opened 2023-01-16
feat: add metrics jitter support

Meemaw:
## About the changes

This PR adds support for `metricsJitter` which can be useful to evenly distribute the load on server for sending metrics.

This is related to https://github.com/Unleash/unleash-proxy/issues/108, and after we applied the jitter to the [python client](https://github.com/Unleash/unleash-client-python/blob/main/UnleashClient/__init__.py#L37) (which already supports this), we saw load on our server even out.

Screenshot 2023-01-11 at 10 34 25

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -6,6 +6,7 @@ import { sdkVersion } from './details.json';
 import { HttpOptions } from './http-options';
 import { suffixSlash } from './url-utils';
 import { UnleashEvents } from './events';
+import { getAppliedJitter } from './helpers';
 
 export interface MetricsOptions {
   appName: string;
@@ -93,8 +94,7 @@ export default class Metrics extends EventEmitter {
   }
 
   private getAppliedJitter() {
-    const jitter = Math.random() * this.metricsJitter;
-    return Math.random() < 0.5 ? -jitter : jitter;
+    return getAppliedJitter(this.metricsJitter);
   }
 
   private startTimer() {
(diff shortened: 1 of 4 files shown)


src/metrics.ts
```diff
@@ -12,8 +12,8 @@ export interface MetricsOptions {
   instanceId: string;
   strategies: string[];
   metricsInterval: number;
+  metricsJitter?: number;
   disableMetrics?: boolean;
-  bucketInterval?: number;
```
Meemaw, line 16: This variable is/was unused.

ivarconr: 👍🏼

ivarconr:
The implementation looks solid. The only question I have is whether there exists a simple way to implement a test for it? 

(I can't think of any not being overly complex)

src/metrics.ts
```diff
@@ -90,13 +92,18 @@ export default class Metrics extends EventEmitter {
     this.httpOptions = httpOptions;
   }
 
+  private getAppliedJitter() {
+    const jitter = Math.random() * this.metricsJitter;
+    return Math.random() < 0.5 ? -jitter : jitter;
+  }
```
FredrikOseberg, line 98: This could be broken out as a unit and tested with a unit test. That would at least be a way to test this functionality in a simple way. Would you mind performing this change?

Meemaw:
@ivarconr @FredrikOseberg I don't think there is much value in testing this. It can't really be tested without mocking pretty much all implementation details (due to randoms). I added a simple test case though I think it provides very little value.

ivarconr (approved):

Issue #405, opened 2023-01-13
feat: resend metrics at next interval if it fails


ivarconr:
### Describe the feature request

Currently, if posting usage metrics back to unleash fails we will drop the metrics and log it as an error. We should instead re-capture the metrics and include them in the next send metrics interval and only log it as a warning.

PR #414, opened 2023-01-16
feat: gracefully handle unsuccessful metrics post

ivarconr:
fixes: #405

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -107,13 +107,13 @@ export default class Metrics extends EventEmitter {
     this.httpOptions = httpOptions;
   }
 
-  private getAppliedJitter() {
+  private getAppliedJitter(): number {
     return getAppliedJitter(this.metricsJitter);
   }
 
-  private startTimer() {
+  private startTimer(): void {
     if (this.disabled) {
-      return false;
+      return;
     }
     this.timer = setTimeout(() => {
       this.sendMetrics();
@@ -122,17 +122,16 @@ export default class Metrics extends EventEmitter {
     if (process.env.NODE_ENV !== 'test' && typeof this.timer.unref === 'function') {
       this.timer.unref();
     }
-    return true;
   }
 
-  start() {
+  start(): void {
     if (typeof this.metricsInterval === 'number' && this.metricsInterval > 0) {
       this.startTimer();
       this.registerInstance();
     }
   }
 
-  stop() {
+  stop(): void {
     if (this.timer) {
       clearInterval(this.timer);
       delete this.timer;
@@ -214,9 +213,9 @@ export default class Metrics extends EventEmitter {
     return true;
   }
 
-  assertBucket(name: string) {
+  assertBucket(name: string): void {
     if (this.disabled) {
-      return false;
+      return;
     }
     if (!this.bucket.toggles[name]) {
       this.bucket.toggles[name] = {
@@ -225,7 +224,6 @@ export default class Metrics extends EventEmitter {
         variants: {},
       };
     }
-    return true;
   }
 
   count(name: string, enabled: boolean): void {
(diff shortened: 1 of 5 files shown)


src/metrics.ts
```diff
@@ -180,107 +194,124 @@ export default class Metrics extends EventEmitter {
...
-      sdkVersion: this.sdkVersion,
-      strategies: this.strategies,
-      started: this.started,
-      interval: this.metricsInterval,
+      bucket,
     };
   }
 
-  getMetricsData(): Data {
+  private reAddBucket(bucket: Bucket) {
```
gardleopard, line 289: I think reAddBucket could be called copyBucket, initializeBucket or something. reAddBucket naming is based on the context of when its called more than what it does.

gardleopard (approved):
LGTM

src/metrics.ts
```diff
@@ -180,107 +194,124 @@ export default class Metrics extends EventEmitter {
...
-    }
+    this.increaseVariantCounter(name, variantName, 1);
 
     this.emit(UnleashEvents.CountVariant, name, variantName);
     return true;
   }
 
-  private bucketIsEmpty() {
-    if (!this.bucket) {
+  private increaseCounter(name: string, enabled: boolean, inc = 1): boolean {
```
sighphyre, line 243: Do we do anything with the return value here or is this to keep in line with the other similar methods?

src/metrics.ts
```diff
@@ -180,107 +194,124 @@ export default class Metrics extends EventEmitter {
...
     }
+    this.assertBucket(name);
+    this.bucket.toggles[name][enabled ? 'yes' : 'no'] += inc;
+    return true;
+  }
+
+  private increaseVariantCounter(name: string, variantName: string, inc = 1): boolean {
+    this.assertBucket(name);
+    if(this.bucket.toggles[name].variants[variantName]) {
+      this.bucket.toggles[name].variants[variantName]+=inc
```
sighphyre, line 255: If you make any changes before merging, would you mind applying an autoformat here? It's not a big deal but it's nice

src/metrics.ts
```diff
@@ -180,107 +194,124 @@ export default class Metrics extends EventEmitter {
...
-      this.bucket.stop = new Date();
-    }
   }
 
-  private getPayload(): Data {
-    this.closeBucket();
-    const payload = this.getMetricsData();
-    this.resetBucket();
-    return payload;
+  private resetBucket(): Bucket {
```
sighphyre, line 274: My gut feel on this function is that it should either mutate state or return a value but not both, it becomes hard to reason about when you see the return value but it's also quietly mutating state behind the scenes

src/url-utils.ts
```diff
@@ -1,4 +1,12 @@
-import { resolve } from 'url';
+export function resolve(from: string, to: string) {
```
sighphyre, line 1: What about `resolveUrl` instead for a name? When it's in the context of `url-utils`, it's easy to know what this does but once its imported and used in another file, it becomes hard to know what this does from the name without looking at its import location

sighphyre (approved):
LGTM, I've left a few thoughts but I don't think anything is deal breaking

src/metrics.ts
ivarconr: lol. these are way old. no real meaning, but was used in tests. will clean up.

ivarconr: fixed

src/metrics.ts
ivarconr: done

src/url-utils.ts
ivarconr: yes. this was a case of me just coping the old deprecated method from url. Agree that resolveUrl is a better name. fixed!

src/metrics.ts
ivarconr: yes, thus ended up as a foot gun. fixing.

PR #415, opened 2023-01-16
Feat/flush metrics

ivarconr:
Adds a flush metrics method, making it easier to gracefully shutdown the SDK and capture the last bucket of metrics. 

This one looks a bit messy because it was built directly on top of #414


sighphyre (approved):
LGTM

PR #417, opened 2023-01-17
fix: flexible-rollout random stickiness is not random enough

ivarconr:
It's easy to get skew when we only pick numbers between 0 and 100.


gardleopard (approved):
LGTM. better with random than radnom as well:D

PR #418, opened 2023-01-18
Allow SDK to startup when backup data is corrupt

sighphyre:
This moves the error that's emitted when the SDK encounters a corrupted JSON backup file to rather be a warning. The backup file shouldn't block the startup sequence, even when it's broken, since the next API hit will cause the SDK to rewrite the file with clean data


nunogois (approved):
LGTM

PR #427, opened 2023-02-17
fix: use base node12 tsconfig

SimenB:
## About the changes

The current transpile target transpiles a bunch of code that's not needed, such as `async` functions and `class`es - both of which are heavily used in this code base.

You target a minimum of node 12, so that's the config I've used: https://github.com/Unleash/unleash-client-node/blob/18ee8ef5a397b8d1615dccd1bbc93edea85bfa0c/.github/workflows/master.yaml#L13

### Important files

`tsconfig.json`

## Discussion points

This is the full diff of the transpiled code: https://gist.github.com/SimenB/6608476192348cc2676dd75e6d76cdf8

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,17 +1,14 @@
 {
+  "extends": "@tsconfig/node12/tsconfig.json",
   "compilerOptions": {
     "rootDir": "./src",
     "outDir": "./lib",
-    "lib": ["es5", "es6"],
-    "target": "es5",
-    "module": "commonjs",
     "sourceMap": true,
     "declaration": true,
     "declarationMap": true,
     "resolveJsonModule": true,
-    "strict": true,
-    "moduleResolution": "node",
-    "downlevelIteration": true
+    "esModuleInterop": false,
+    "skipLibCheck": false
   },
   "exclude": ["examples/", "lib/", "node_modules/"]
 }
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -1,17 +1,14 @@
 {
+  "extends": "@tsconfig/node12/tsconfig.json",
   "compilerOptions": {
     "rootDir": "./src",
     "outDir": "./lib",
-    "lib": ["es5", "es6"],
-    "target": "es5",
-    "module": "commonjs",
```
SimenB, line 7: comes from extended config

tsconfig.json
```diff
@@ -1,17 +1,14 @@
...
     "sourceMap": true,
     "declaration": true,
     "declarationMap": true,
     "resolveJsonModule": true,
-    "strict": true,
-    "moduleResolution": "node",
```
SimenB, line 13: comes from extended config

tsconfig.json
```diff
@@ -1,17 +1,14 @@
...
-    "downlevelIteration": true
```
SimenB, line 14: why?

tsconfig.json
```diff
@@ -1,17 +1,14 @@
...
+    "esModuleInterop": false,
+    "skipLibCheck": false
```
SimenB, line 11: overrides extended config to be stricter (and matches what you had before)

ivarconr (approved):
Looks good.  Thanks 🙏👍

PR #428, opened 2023-02-17
chore: run CI on node versions 18 and 19

SimenB:
## About the changes

Node 18 is the current LTS and Node 19 is the current version. Verifying compatibility seems reasonable

Closes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,16 +9,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [14.x]
-
     steps:
     - uses: actions/checkout@v3
     - name: Publish to npm
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: lts/*
         registry-url: 'https://registry.npmjs.org'
     - run: |
         yarn install --frozen-lockfile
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -9,21 +9,17 @@ jobs:
...
-      matrix:
-        node-version: [14.x]
-
     steps:
     - uses: actions/checkout@v3
     - name: Publish to npm
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: lts/*
```
SimenB, line 17: this will currently resolve to v18. Not sure if you wanna lock it down or not? Either way, a matrix is not needed here

test/repository.test.js
```diff
@@ -300,7 +300,8 @@ test('should handle invalid JSON response', t =>
       t.truthy(err);
       t.true(
         err.message.indexOf('Unexpected token') > -1 ||
-                    err.message.indexOf('Unexpected end of JSON input') > -1,
+                    err.message.indexOf('Unexpected end of JSON input') > -1 ||
+                    err.message.indexOf(' Unterminated string in JSON') > -1,
```
SimenB, line 304:
```suggestion
                    err.message.indexOf('Unterminated string in JSON') > -1,
```

Tymek (approved):

PR #429, opened 2023-02-17
chore: use builtin caching of `actions/setup-node`

SimenB:
## About the changes

The current action is spewing out deprecation warnings.

image

`actions/setup-node` has had builtin `node_modules` caching for a long time at this point, so just using that seems best

(coveralls is https://github.com/coverallsapp/github-action/issues/132)

Closes #

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -19,8 +19,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{matrix.node-version}}
-      - name: Setup yarn cache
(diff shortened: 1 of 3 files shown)


thomasheartman (approved):
Yeah, this seems like a pretty straightforward upgrade to me. Can't see any reason not to do it 💯 

But if I understood your comment about coveralls correctly: there's no fix/upgrade path for that just yet, right?

SimenB:
yep. I use codecov in my own projects, might be something to consider here to move off of an unmaintained module 🙂

PR #434, opened 2023-03-02
Fix: Add check on impressionData

dev-dennis-guye:
## About the changes

[Impression Data](https://docs.getunleash.io/reference/impression-data) is an opt-in feature. However, the previous implementation triggered events regardless of the status of the impressionData field. This fix adds a check on the status of the impressionData field on FeatureInterface before triggering the event.

### Important files

src/client.ts

## Discussion points

This fix addresses that issue however is more invasive than I would prefer. The change required making an update to the interface of the resolve variant function.

### DIFF
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -4,13 +4,15 @@ import { Strategy } from '../lib/strategy';
 
 import CustomStrategy from './true_custom_strategy';
 import CustomFalseStrategy from './false_custom_strategy';
+import { UnleashEvents } from '../lib/events';
 
-function buildToggle(name, active, strategies, variants = []) {
+function buildToggle(name, active, strategies, variants = [], impressionData= false) {
   return {
     name,
     enabled: active,
     strategies: strategies || [{ name: 'default' }],
     variants,
+    impressionData
   };
 }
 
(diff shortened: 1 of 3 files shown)


FredrikOseberg (approved):
LGTM, could you run lint:fix to get the correct formatting. Also, the tests are currently timing out in CI.

dev-dennis-guye:
> LGTM, could you run lint:fix to get the correct formatting. Also, the tests are currently timing out in CI.

So I took a look, my hypothesis is that this was caused by 

```typescript
client.on(UnleashEvents.Impression, ()=> {
    t.pass()
  });
```

My guess is that Ava expects t.pass() to be the last function executed. I've reimplemnted the tests to instead perform

```typescript
let called=false; 
//code
client.on(UnleashEvents.Impression, ()=> {
    called=true;
  });
//code 
t.true(called); 
```

dev-dennis-guye:
@FredrikOseberg Any update on this?

FredrikOseberg:
Hey, thanks for the prompt. I will have another look tomorrow

FredrikOseberg:
> Details

@dev-dennis-guye Hm. Looks like the tests are timing out here in CI.

Tymek (approved):

test/client.test.js
```diff
@@ -4,13 +4,15 @@ import { Strategy } from '../lib/strategy';
 
 import CustomStrategy from './true_custom_strategy';
 import CustomFalseStrategy from './false_custom_strategy';
+import { UnleashEvents } from '../lib/events';
 
-function buildToggle(name, active, strategies, variants = []) {
+function buildToggle(name, active, strategies, variants = [], impressionData= false) {
```
Tymek, line 9: Looks like a missing space before `=`. I don't know if prettier is working correctly.

Tymek (changes requested):

test/client.test.js
dev-dennis-guye, line 9:
```suggestion
function buildToggle(name, active, strategies, variants = [], impressionData = false) {
```

I'm assuming this is the correct formatting.

FredrikOseberg:
@dev-dennis-guye It looks like your tests are timing out in CI. You might need to make this an async function and use promises to make sure your callback is being called.

Skjermbilde 2023-03-13 kl  08 53 41

FredrikOseberg:
@dev-dennis-guye You still around?

FredrikOseberg:
> > @dev-dennis-guye You still around?
> 
> @FredrikOseberg Sorry for the late update. This is is not a promise issue and is also unrelated to the listener remaining attached at the end of the test.

Yeah. My guess is that the tests that were testing this functionality is now hanging because of the added check.

FredrikOseberg (approved):
LGTM

PR #442, opened 2023-04-07
fix: always resolve startUnleash promise

### LINKED ISSUE #441: startUnleash functions doesn't return when called multiple times
### Describe the bug

`startUnleash` is supposed to return a singleton. Per definition, it should return the same instance when invoked multiple times. What happens, instead, is that it works only when invoked the first time. From the second on, the instance is not returned.

Use the repro repo at https://github.com/fox1t/unleash-bug
Instead of logging two console logs, only the first one is visible.

### Expected behavior

The `startUnleash` invoked function returns the instance correctly when called multiple times.

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

A possible fix is to emit the `synchronized` event inside the `getIntance` method.
```ts
if (Unleash.instance) {
	if (configSignature !== Unleash.configSignature) {
		throw new Error(
			"You already have an Unleash instance with a different configuration.",
		);
	}
	process.nextTick(() => Unleash.instance.emit("synchronized"));
	return Unleash.instance;
}
```

### Unleash version

4.22.1

### Subscription type

Pro

### Hosting type

Hosted by Unleash

3.19.0

fooddilsn:
## About the changes

`startUnleash` should wait for the `synchronized` event only when the Unleash instance isn't already synchronized.
Without the changes introduced by this PR, any subsequent invocation of `startUnleash` will produce an unresolved promise.

Closes #441

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,7 +24,9 @@ export function initialize(options: UnleashConfig): Unleash {
 
 export async function startUnleash(options: UnleashConfig): Promise<Unleash> {
   const unleash = initialize(options);
-  await once(unleash, 'synchronized');
+  if (!unleash.isSynchronized()) {
+    await once(unleash, 'synchronized');
+  }
   return unleash;
 }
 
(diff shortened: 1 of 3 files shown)


ivarconr (approved):

ivarconr:
Thank you @fooddilsn

PR #452, opened 2023-05-16
fix: specify all payload types and export them

nunogois:
Is there any reason why we're not specifying all variant payload types and exporting them?

My use case is leveraging this in [unleash](https://github.com/unleash/unleash) so I can have proper types and type-checking.

image


sighphyre (approved):
I don't believe there's a reason for not doing this. Seems like a reasonable change

LGTM!

PR #453, opened 2023-05-16
3.19.2

nunogois:
Bumps to `3.19.2` to cut another patch release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "unleash-client",
-  "version": "3.19.1",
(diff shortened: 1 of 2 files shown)


sighphyre:
We changed the public API here so would a 3.20.0 not be more appropriate?

nunogois:
> We changed the public API here so would a 3.20.0 not be more appropriate?

Yeah that's fair. Opened https://github.com/Unleash/unleash-client-node/pull/454

PR #455, opened 2023-05-16
fix: getToggle by name may not find the toggle

gastonfournier:
## About the changes
Node SDK interface should notify that getting a toggle by name can return undefined if the toggle does not exist:
https://github.com/Unleash/unleash/blob/d298b04cb2864b4853ca456e402714c98adaf932/src/lib/proxy/proxy-repository.ts#L75-L77

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -70,7 +70,11 @@ export default class UnleashClient extends EventEmitter {
     return enabled;
   }
 
-  isFeatureEnabled(feature: FeatureInterface, context: Context, fallback: Function): boolean {
+  isFeatureEnabled(
+    feature: FeatureInterface | undefined,
+    context: Context,
+    fallback: Function,
+  ): boolean {
     if (!feature) {
       return fallback();
     }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -70,7 +70,11 @@ export default class UnleashClient extends EventEmitter {
     return enabled;
   }
 
-  isFeatureEnabled(feature: FeatureInterface, context: Context, fallback: Function): boolean {
+  isFeatureEnabled(
+    feature: FeatureInterface | undefined,
+    context: Context,
+    fallback: Function,
+  ): boolean {
```
thomasheartman, line 77: Why are we changing the accepted parameter value? Surely, there's no point in ever calling this with a feature that's `undefined`? Wouldn't it be better to only allow you to call it with a defined feature?

thomasheartman (approved):
Looks like a sensible change to me: I love that we're explicit about returning undefined if the feature doesn't exist 😄  But I've got a question on relaxing function parameters: I'm not sure we should allow `isFeatureEnabled` to be called with an `undefined` feature. 🤔 But that said, I'm guessing we already do that today, so this PR only updates the code to reflect reality and does not do anything to address any issues?

src/client.ts
gastonfournier: yes! Very good point, this is just the result of type refactoring, but I can look into this as it definitely makes sense!

sighphyre (approved):
Spec tests pass, change looks reasonable LGTM

ivarconr (approved):
Nice. I consider this a bug-fix.

PR #460, opened 2023-06-07
Update workflows

Tymek:
Unified release action


gastonfournier (approved):
This looks nice! Is this generating also the changelog? I notice this also handles the version, so this could be a nice improvement for our release process in Unleash

Tymek:
> Is this generating also the changelog?

Yes. https://github.com/Unleash/.github/blob/main/.github/workflows/npm-release.yml#L74-L94

PR #482, opened 2023-07-10
feat: strategy variants

kwasniew:
## About the changes

Adding support for strategy variants.

Screenshot 2023-07-12 at 16 15 18

Background: Unleash supports feature environment variants. This PR adds variants capability per strategy which is more flexible and easier to understand/implement. It makes it possible to use constraints and segments for variants.

How it's implemented?
If a strategy has a corresponding variant param it's used over the feature environment variant. 

## Discussion points

I'm thinking if we should have env var or config param to enable this new capability in addition to variant params detection.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -105,12 +105,17 @@ export default class UnleashClient extends EventEmitter {
           return false;
         }
         const constraints = this.yieldConstraintsFor(strategySelector);
-        const enabled = strategy.isEnabledWithConstraints(strategySelector.parameters, context, constraints);
-        if(enabled && strategySelector?.parameters?.variant) {
-          if(typeof strategySelector?.parameters?.variant === 'string') {
-            featureVariant = {name: strategySelector?.parameters?.variant, payload: {type: 'string', value: strategySelector?.parameters?.variant}}
+        const enabled =
+          strategy.isEnabledWithConstraints(strategySelector.parameters, context, constraints);
+        const variantParam = strategySelector?.parameters?.variant;
+        if (enabled && variantParam) {
+          if (typeof variantParam === 'string') {
+            featureVariant = {
+              name: variantParam,
+              payload: { type: 'string', value: variantParam },
+            };
           } else {
-            featureVariant = strategySelector?.parameters?.variant;
+            featureVariant = variantParam;
           }
         }
         return enabled;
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -74,40 +74,56 @@ export default class UnleashClient extends EventEmitter {
     feature: FeatureInterface | undefined,
     context: Context,
     fallback: Function,
-  ): boolean {
+  ): [boolean, VariantDefinition | undefined] {
```
kwasniew, line 77: in addition to boolean true/false was want to know the actual variant that the strategy resolved to if it was provided.

src/client.ts
```diff
@@ -74,40 +74,56 @@ export default class UnleashClient extends EventEmitter {
...
           this.warnOnce(strategySelector.name, feature.name, feature.strategies);
           return false;
         }
         const constraints = this.yieldConstraintsFor(strategySelector);
-        return strategy.isEnabledWithConstraints(strategySelector.parameters, context, constraints);
+        const enabled =
+          strategy.isEnabledWithConstraints(strategySelector.parameters, context, constraints);
+        const variantParam = strategySelector?.parameters?.variant;
+        if (enabled && variantParam) {
+          if (typeof variantParam === 'string') {
```
kwasniew, line 112: we support simple string based variants

src/client.ts
```diff
@@ -74,40 +74,56 @@ export default class UnleashClient extends EventEmitter {
...
+            featureVariant = {
+              name: variantParam,
+              payload: { type: 'string', value: variantParam },
+            };
+          } else {
+            featureVariant = variantParam;
```
kwasniew, line 118: and full blown variants for json/csv/number support

src/client.ts
```diff
@@ -169,25 +185,33 @@ export default class UnleashClient extends EventEmitter {
...
 
     let enabled = true;
+    let strategyVariant = undefined;
     if (checkToggle) {
-      enabled = this.isFeatureEnabled(feature, context, () =>
+      [enabled, strategyVariant] = this.isFeatureEnabled(feature, context, () =>
         fallbackVariant ? fallbackVariant.enabled : false,
       );
-      if (!enabled) {
+      if (!enabled ||
```
kwasniew, line 199: we moved early exit here

Tymek (approved):

Tymek:
Should should bump SDK compatibility in the future? It will allow us to add warnings if you're using old SDKs with new Variants. https://github.com/Unleash/unleash-client-node/blob/b30488f866acd6160de78f2f790466eba41fbe43/src/repository/index.ts#L13

sighphyre (approved):
Okay wow, that's cool, spec tests pass and all!

sighphyre (approved):
LGTM

PR #489, opened 2023-07-14
fix: remove unused segment fields

kwasniew:
## About the changes

Interface for segments was incorrect with unnecessary fields

### DIFF
--- a/src/strategy/strategy.ts
+++ b/src/strategy/strategy.ts
@@ -23,10 +23,7 @@ export interface Constraint {
 export interface Segment {
   id: number;
   name: string;
-  description?: string;
   constraints: Constraint[];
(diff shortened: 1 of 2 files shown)


src/strategy/strategy.ts
```diff
@@ -23,10 +23,7 @@ export interface Constraint {
 export interface Segment {
   id: number;
   name: string;
```
sighphyre, line 25: Do we use name anywhere?

kwasniew: I can remove it too. I left it in the backend response for now though. But maybe I can remove it too.

sighphyre (approved):
Awesome, nice change!

PR #491, opened 2023-07-18
feat: expose config type

kwasniew:
## About the changes

Expose UnleashConfig type for the Unleash Proxy to use


sighphyre (approved):
Fair enough! LGTM!

PR #493, opened 2023-07-19
fix: force get variant is back

kwasniew:
## About the changes

forceGetVariant should still be used in proxy/frontend API because it's not emitting impression events. The only difference is that it now evaluates the toggle under the hood (for the strategy variants support) because the experiments confirmed it's not skewing % distribution. I'm afraid that if we switch proxy and frontend SDK to getVariant the impression event will be emitted twice from the client SDK AND frontend API/proxy

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "unleash-client",
-  "version": "4.1.0-beta.2",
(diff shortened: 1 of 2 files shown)


thomasheartman (dismissed):
Looks nice 😄

PR #494, opened 2023-07-19
refactor: strategy variants inside strategy

kwasniew:
## About the changes

* explicitly modelling strategy result as a discriminative union. We're discriminating on the enabled key.
* strategy variant calculation moved to the strategy instead of being spread between client and variant
* using strategy groupId for variant distribution instead of using feature name. groupId can be set to the feature name though. This change allows to set the same groupId across variants in different features strategies

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4,10 +4,10 @@ import { FeatureInterface } from './feature';
 import { RepositoryInterface } from './repository';
 import {
   Variant, VariantDefinition,
-  getDefaultVariant, selectVariant, selectVariantDefinition,
+  getDefaultVariant, selectVariant,
 } from './variant';
 import { Context } from './context';
-import { Constraint, Segment } from './strategy/strategy';
+import { Constraint, Segment, StrategyResult } from './strategy/strategy';
 import { createImpressionEvent, UnleashEvents } from './events';
 
 interface BooleanMap {
@@ -58,7 +58,7 @@ export default class UnleashClient extends EventEmitter {
 
   isEnabled(name: string, context: Context, fallback: Function): boolean {
     const feature = this.repository.getToggle(name);
-    const [enabled] = this.isFeatureEnabled(feature, context, fallback);
+    const { enabled } = this.isFeatureEnabled(feature, context, fallback);
     if (feature?.impressionData) {
       this.emit(
         UnleashEvents.Impression,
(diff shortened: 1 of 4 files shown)


src/variant.ts
```diff
@@ -77,7 +77,7 @@ function findOverride(variants: VariantDefinition[],
 }
 
 export function selectVariantDefinition(
-  featureName: string,
+  groupId: string,
```
kwasniew, line 80: featureName was misleading in the original code since we accept groupId here

src/client.ts
```diff
@@ -77,47 +77,48 @@ export default class UnleashClient extends EventEmitter {
...
-      return [feature.enabled, undefined];
+      return { enabled: feature.enabled };
     }
 
-    let featureVariant = undefined;
+    let strategyResult = { enabled: false };
 
-    return [(
-      feature.strategies.length > 0 &&
+    feature.strategies?.some((strategySelector): boolean => {
```
kwasniew, line 101: I left some in the impl since we want the early exit/short-circuit semantics

src/strategy/strategy.ts
```diff
@@ -215,4 +218,31 @@ export class Strategy {
   ) {
     return this.checkConstraints(context, constraints) && this.isEnabled(parameters, context);
   }
+
+  getResult(parameters: any,
```
kwasniew, line 222: this is the main change - everything strategy variant related is located in one class

src/client.ts
thomasheartman: Probably not super relevant, but is `some` guaranteed to traverse the list in order? Or is it just ... most likely to do that?

src/strategy/strategy.ts
```diff
@@ -166,6 +166,9 @@ operators.set(Operator.DATE_BEFORE, DateOperator);
 operators.set(Operator.SEMVER_EQ, SemverOperator);
 operators.set(Operator.SEMVER_GT, SemverOperator);
 operators.set(Operator.SEMVER_LT, SemverOperator);
+
+export type StrategyResult = { enabled: true, variant?: Variant | null } | { enabled: false };
```
thomasheartman, line 170: Why is variant both undefineable and null? Do we treat them differently? That seems subtle and like something that might cause issues down the line. Why not just undefined?

```suggestion
export type StrategyResult = { enabled: true, variant?: Variant } | { enabled: false };
```

thomasheartman (approved):
Looks sensible to me. I've got two questions that I'd like to get answered, but nothing super major. 

Assuming using the group ID over feature name is good, I'm giving this a provisional approval.

src/strategy/strategy.ts
kwasniew: One of the methods that we call (selecting variant from a list) returns null so we have to handle this

kwasniew:
@thomasheartman Array.some in practice gives me predictable order but I don't think spec can guarantee this. We should have tests that check this.

src/strategy/strategy.ts
thomasheartman: Ah. In that case, could we just make it `Variant | null`? Because we allow both it strikes me as significant, so  I would expect `undefined` and `null` to mean different things. If they don't, I would prefer collapsing it into a single value.

Your choice in the end, but the more we can lock it down and be strict about what we return, the easier it'll be to work with later.

thomasheartman:
@kwasniew Regarding guaranteed order with some: Arrays are ordered collections, so it should be fine. I'm just thinking that some engines might perform some optimizations one the implementation that makes it not predictable (because the function itself doesn't really care about order). But looking back at the code: the order here really doesn't matter, does it? So whether the order is guaranteed or not is irrelevant, yeah? Never mind me, then 😅

sighphyre:
> @thomasheartman Array.some in practice gives me predictable order but I don't think spec can guarantee this. We should have tests that check this.

It's actually guaranteed to be ordered in the language spec!

https://262.ecma-international.org/5.1/#sec-15.4.4.17

kwasniew:
@sighphyre good call. I should have checked directly in the spec instead of asking chatgpt :D 
For future readers of the PR: "some calls callbackfn once for each element present in the array, in ascending order,"

src/strategy/strategy.ts
kwasniew: After further investigation we can remove null

PR #495, opened 2023-07-20
feat: variant on enabled toggle metrics

kwasniew:
## About the changes

* If toggle is enabled and variant doesn't exist, getVariant should still count enabled toggle as yes
* Unit tests show this new behavior
* This change makes the bahavior same as JS SDK and Java SDK


src/unleash.ts
```diff
@@ -302,18 +306,21 @@ export class Unleash extends EventEmitter {
...
       this.emit(
         UnleashEvents.Warn,
         `Unleash has not been initialized yet. isEnabled(${name}) defaulted to ${result}`,
       );
     }
+    const { featureEnabled, ...variant } = result;
     if (result.name) {
       this.countVariant(name, result.name);
     }
-    this.count(name, result.enabled);
```
kwasniew, line 314: we were returning variant.enabled not feature enabled

Tymek (approved):

src/test/snapshots/client.test.ts.md
```diff
@@ -10,6 +10,7 @@ Generated by [AVA](https://avajs.dev).
 
     {
       enabled: true,
+      featureEnabled: true,
```
gastonfournier, line 13: there was a value `enabled` previously. Doesn't it correspond to the feature being enabled? I don't think we have a variant enabled/disabled status

kwasniew: If that's the case I could massively simplify the fix

kwasniew: the enabled field is there but I thought it's for the variant not the toggle status

src/unleash.ts
sighphyre: Is this the result of `isEnabled` or is this the raw enabled field that comes back from the Unleash API?

kwasniew: it wasn't, it was the enabled status of the variant itself.

kwasniew: In other SDKs we have isEnabled result in scope so we just use it. Here the isEnabled is called 2 layers deep so we need to propagate it up the stack

kwasniew: This PR is doing the following - for enabled and user activated toggle with no variants:
client.getVariant() returns disabled variant and reports enabled toggle metric

sighphyre: Thanks! Makes sense!

sighphyre (approved):
Ye, I think this makes sense

PR #496, opened 2023-07-24
feat: version bump non beta

kwasniew:
## About the changes

We were testing this version ourselves over the weekend with no issues, time to make it non-beta


sighphyre (approved):
WOOT!

PR #520, opened 2023-09-19
feat: dependent features

kwasniew:
## About the changes

Dependent features implementation based on spec: https://github.com/Unleash/client-specification/pull/63

Extra checks that may also be covered by other SDKs (listing them since this is our reference implementation):
* metrics are not called on dependent features
* impression events are called on dependent features for easier debugging
* warning events for missing dependencies are reported once

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -19,13 +19,16 @@ export default class UnleashClient extends EventEmitter {
 
   private strategies: Strategy[];
 
-  private warned: BooleanMap;
+  private warnedStrategies: BooleanMap;
+
+  private warnedDependencies: BooleanMap;
 
   constructor(repository: RepositoryInterface, strategies: Strategy[]) {
     super();
     this.repository = repository;
     this.strategies = strategies || [];
-    this.warned = {};
+    this.warnedStrategies = {};
+    this.warnedDependencies = {};
 
     this.strategies.forEach((strategy: Strategy) => {
       if (
@@ -43,14 +46,26 @@ export default class UnleashClient extends EventEmitter {
     return this.strategies.find((strategy: Strategy): boolean => strategy.name === name);
   }
 
-  warnOnce(missingStrategy: string, name: string, strategies: StrategyTransportInterface[]) {
-    if (!this.warned[missingStrategy + name]) {
-      this.warned[missingStrategy + name] = true;
+  warnStrategyOnce(missingStrategy: string,
+                   name: string,
+                   strategies: StrategyTransportInterface[]) {
+    if (!this.warnedStrategies[missingStrategy + name]) {
+      this.warnedStrategies[missingStrategy + name] = true;
       this.emit(
         UnleashEvents.Warn,
         `Missing strategy "${missingStrategy}" for toggle "${name}". Ensure that "${strategies
           .map(({ name: n }) => n)
           .join(', ')}" are supported before using this toggle`,
+      );
+    }
+  }
+
+  warnDependencyOnce(missingDependency: string, name: string) {
+    if (!this.warnedDependencies[missingDependency + name]) {
+      this.warnedDependencies[missingDependency + name] = true;
+      this.emit(
+        UnleashEvents.Warn,
+        `Missing dependency "${missingDependency}" for toggle "${name}"`,
       );
     }
   }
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -55,8 +55,37 @@ export default class UnleashClient extends EventEmitter {
     }
   }
 
+  isParentDependencySatisfied(feature: FeatureInterface | undefined, context: Context) {
```
sjaanus, line 58: Why undefined?

src/client.ts
```diff
@@ -55,8 +55,37 @@ export default class UnleashClient extends EventEmitter {
...
+    if (!feature?.name) {
```
sjaanus, line 59: Can feature not have name?

src/client.ts
```diff
@@ -55,8 +55,37 @@ export default class UnleashClient extends EventEmitter {
...
+
+        if (parentFeature?.dependencies?.length) {
+          return false;
+        }
+
+        if (parent.enabled !== false && parent.variants?.length) {
+          return parent.variants.includes(this.getVariant(parent.feature, context).name);
+        }
+
+        return parent.enabled !== false
```
sjaanus, line 75: Can we make it more readable, and not do negate operator, but instead `===`

src/client.ts
```diff
@@ -55,8 +55,37 @@ export default class UnleashClient extends EventEmitter {
...
+  isParentDependencySatisfied(feature: FeatureInterface | undefined, context: Context) {
+    if (!feature?.name) {
+      return true;
+    }
+
+    if (feature.dependencies && feature.dependencies.length) {
+      return feature.dependencies.every(parent => {
+        const parentFeature = this.repository.getToggle(parent.feature);
+
+        if (parentFeature?.dependencies?.length) {
```
sjaanus, line 67: Can we make it more readable, Does this mean length exists or length === 0?

src/client.ts
kwasniew: by default enabled is not provided so it's undefined. enabled: false is a special case for kill-switches in parent flag.This code is explicit about explicit enabled false and everything else.

src/client.ts
kwasniew: this is what the existing type was telling me so I'm respecting it here

src/client.ts
kwasniew: existing types indicated so

src/client.ts
kwasniew: nope, it should also cater for undefined dependencies. Setting === 0 breaks the spec

src/client.ts
kwasniew: I reorganized the conditions a bit but the !== false has to stay as it is

src/client.ts
kwasniew: move it to precondition check though

src/client.ts
```diff
@@ -55,8 +55,32 @@ export default class UnleashClient extends EventEmitter {
...
+        return this.isEnabled(parent.feature, context, () => false);
+      }
+
+      return !this.isEnabled(parent.feature, context, () => false);
+    });
+  }
+
   isEnabled(name: string, context: Context, fallback: Function): boolean {
     const feature = this.repository.getToggle(name);
+    if(!this.isParentDependencySatisfied(feature, context)) {
```
sighphyre, line 81: Doesn't this guard clause mean there's no impression data for the toggle if the parent causes it to be disabled?

src/client.ts
```diff
@@ -55,8 +55,32 @@ export default class UnleashClient extends EventEmitter {
     }
   }
 
+  isParentDependencySatisfied(feature: FeatureInterface | undefined, context: Context) {
+    if (!feature?.name || !feature.dependencies?.length) {
```
sighphyre, line 59: What's with the name check here? Name should always be a non empty string, do I have the dumb or is the same as 

if (!feature?.dependencies?.length) {
  return true;
}

kwasniew: still passes the spec. not sure why I did the name check but it's not in the type anymore. I'm hallucinating :D

src/client.ts
kwasniew: Done

src/client.ts
kwasniew: Removed

src/client.ts
kwasniew: added a test to make sure no one makes the same mistake again :)

sighphyre (commented):
Glancing over the metrics code, it looks correct but I think it's worth having one or two tests to check that the parent doesn't get metric counts (I'm guessing, don't know if that's what you want). Especially if this is going to be the reference implementation so others know what to do, since the spec doesn't handle that

kwasniew:
@sighphyre addressed all your points :)

src/client.ts
sighphyre: Feature's should have a name, but at least in this SDK we allow people to define their own bootstrapper so it's possible people put goop in there. This one is probably mostly legacy but respecting the existing types seems very sane to me

sighphyre (approved):
Nice! LGTM

ivarconr (dismissed):

PR #530, opened 2023-10-18
feat: force get variant with strategy variants

kwasniew:
## About the changes

We'd like to avoid double evaluation of strategies when making 2 separate calls to check enabled status and get variant. 

* Adding new `evaluate` method that returns full StrategyResult with enabled status and optional variant. 
* forceGetVariant accepts optional StrategyResult parameter with a forced value
* as a bonus: moving parent dependency checking to one place 

We follow this pattern in playground. Now it will be possible to fix frontend API and proxy:
```ts
 return definitions
            .map((feature) => [feature, client.evaluate(feature.name, context)] as const)
            .filter(([,result]) => result.enabled)
            .map(([feature, result]) => ({
                name: feature.name,
                enabled: Boolean(feature.enabled),
                variant: client.forceGetVariant(feature.name, context, undefined, result),
                impressionData: Boolean(feature.impressionData),
            }));
```
In the code above we call evaluate to get enabled info + possible strategy variant. We can then pass the strategy variant to forceGetVariant to avoid double evaluation. 

This change has 2 new unit tests + was tested in the frontend API.

### DIFF
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -323,7 +323,10 @@ export class Unleash extends EventEmitter {
     return variant;
   }
 
-  forceGetVariant(name: string, context: Context = {}, fallbackVariant?: Variant, forcedResult?: StrategyResult): Variant {
+  forceGetVariant(name: string,
+                  context: Context = {},
+                  fallbackVariant?: Variant,
+                  forcedResult?: StrategyResult): Variant {
     const enhancedContext = { ...this.staticContext, ...context };
     let result;
     if (this.ready) {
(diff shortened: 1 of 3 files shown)


src/unleash.ts
```diff
@@ -276,6 +277,28 @@ export class Unleash extends EventEmitter {
     return result;
   }
 
+  evaluate(name: string,
```
kwasniew, line 280: new method added to our public API. This is the same is isEnabled but adds extra info about strategy evaluation result

kwasniew: the intended user of this method will be proxy and frontend API

PR #531, opened 2023-10-18
refactor: one parent dependency check

kwasniew:
## About the changes

Reduced number of calls to parent dependency check.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -99,13 +99,7 @@ export default class UnleashClient extends EventEmitter {
 
   isEnabled(name: string, context: Context, fallback: Function): boolean {
     const feature = this.repository.getToggle(name);
-    let enabled: boolean;
-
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -131,7 +125,7 @@ export default class UnleashClient extends EventEmitter {
       return { enabled: fallback() };
     }
 
-    if (!feature || !feature.enabled) {
+    if (!feature || !this.isParentDependencySatisfied(feature, context) || !feature.enabled) {
```
kwasniew, line 128: by moving parent dependency check to isFeatureEnabled we can simplify isEnabled method and resolveVariant since both of those methods call isFeatureEnabled under the hood.

Tymek, line 128:
```suggestion
    if (!feature || !feature.enabled || !this.isParentDependencySatisfied(feature, context)) {
```

Tymek (approved):

PR #536, opened 2023-11-09
feat: variant feature status - client specification update

Tymek:
## About the changes

Adds a new "feature_enabled" property to the variant response, similar to https://github.com/Unleash/unleash-client-python/pull/285

https://linear.app/unleash/issue/SR-152/node-sdk-add-feature-enabled-property-to-variant-response


src/variant.ts
```diff
@@ -32,18 +32,14 @@ export interface Variant {
   name: string;
   enabled: boolean;
   payload?: Payload;
+  feature_enabled?: boolean;
```
nunogois, line 35: Do we need this one to be optional here? 🤔 Would it break anything if we made it required?

nunogois (approved):
LGTM, just the one question.

src/variant.ts
Tymek: We can make it required in next version. I'd like to avoid breaking "fallback variant" interface

PR #537, opened 2023-11-15
feat: Backoff based on HTTP statuses

chriswk:
### What
This PR adds incremental backoff to both metrics posting as well as feature fetching.
It does so by keeping track of the number of failures we've seen and adding `interval + (failureCount * interval)` to the next `setTimeout` call. Each succeeding call of the same type will decrement failure count, reducing the interval respectively.

### Discussion points.
I've set maximum failure count to 10 to ensure we never wait forever before the next poll/post of metrics.


src/metrics.ts
```diff
@@ -111,21 +114,39 @@ export default class Metrics extends EventEmitter {
...
+      this.getInterval(),
+    );
 
     if (process.env.NODE_ENV !== 'test' && typeof this.timer.unref === 'function') {
       this.timer.unref();
     }
   }
 
   start(): void {
-    if (typeof this.metricsInterval === 'number' && this.metricsInterval > 0) {
```
chriswk, line 128: Typescript compiler stated that this was no longer necessary

src/math.ts
```diff
@@ -0,0 +1,7 @@
+export function max(a: number, b: number): number {
```
sighphyre, line 1: Do these do something different from `Math.max`? Fiddling in the browser console, they kinda look like the same thing

src/repository/index.ts
```diff
@@ -89,6 +95,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
     instanceId,
     projectName,
     refreshInterval = 15_000,
+    maxRefreshInterval = 120_000,
```
sighphyre, line 98: If we expose this in the constructor, it gives consumers the ability to just turn off our backoff. Do we want that? Tempted to use a constant here

yarn.lock
```diff
@@ -812,18 +812,16 @@ acorn@^8.9.0:
 
 agent-base@6, agent-base@^6.0.2:
   version "6.0.2"
-  resolved "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz"
+  resolved "https://registry.yarnpkg.com/agent-base/-/agent-base-6.0.2.tgz#49fff58577cfee3f37176feab4c22e00f86d7f77"
```
sighphyre, line 815: The yarn lock update needed or just boyscouting?

Neither is an issue, just curious for context

src/repository/index.ts
```diff
@@ -241,12 +249,31 @@ Message: ${err.message}`,
...
+    this.failures = min(this.failures + 1, 10);
+    return this.nextFetch();
+  }
+
+  private countSuccess(): number {
+    this.failures = max(this.failures - 1, 0);
+    return this.nextFetch();
+  }
+
   async fetch(): Promise<void> {
```
sighphyre, line 270: `fetch` is now just short of 100 lines. Might be worth breaking this down into smaller methods at this point

sighphyre (approved):
Looks like it does what it says on the tin, tests seem to cover it nicely too

Suggested some small things but I'll leave the choices up to you

src/math.ts
chriswk: Oh, just not aware of Math.min, Math.max, I was trying to find it in the default library and my importer couldn't

chriswk: Thanks

chriswk: Fixed in 799a51c

yarn.lock
chriswk: This, happened automatically when I ran yarn install.

src/repository/index.ts
chriswk: agreed. Something like what we did for metrics. I'll have a look.

src/repository/index.ts
chriswk: It's removed, and replaced by putting a maximum backoff of 10 * interval

nunogois (dismissed):
LGTM

src/metrics.ts
```diff
@@ -170,6 +190,19 @@ export default class Metrics extends EventEmitter {
     return true;
   }
 
+  unrecoverableError(url: string, statusCode: number) {
+    this.emit(UnleashEvents.Warn, `${url} returning ${statusCode}, stopping metrics`);
+    this.failures = -1;
```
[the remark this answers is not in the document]
daveleek: I can see the benefit in having a value indicate unrecoverable error, but I think it should in that case be a separate variable?

src/metrics.ts
```diff
@@ -194,16 +227,22 @@ export default class Metrics extends EventEmitter {
...
         httpOptions: this.httpOptions,
       });
-      this.startTimer();
-      if (res.status === 404) {
-        this.emit(UnleashEvents.Warn, `${url} returning 404, stopping metrics`);
-        this.stop();
-      }
       if (!res.ok) {
+        if (res.status === 404 || res.status === 403 || res.status == 401) {
+          this.unrecoverableError(url, res.status);
```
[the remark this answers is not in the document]
daveleek: I remember from the quick discovery/discussion session we had that we thought maybe not stop polling just treat it as a backoff-response, for the reasons @gastonfournier mentions here

daveleek (approved):
Looks good! I noticed the implementations are a little bit different between the metrics and the features pollings. Is that intentional?

src/metrics.ts
```diff
@@ -111,21 +113,39 @@ export default class Metrics extends EventEmitter {
     return getAppliedJitter(this.metricsJitter);
   }
 
+  getFailures(): number {
+    return this.failures;
+  }
+
+  getInterval(): number {
+    if(this.metricsInterval === 0 || this.failures === -1) {
```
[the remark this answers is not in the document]
chriswk: failures start at 0, if we initialize it to -1 we'd never actually try to post metrics.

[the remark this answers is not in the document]
chriswk: But since you're asking, I guess it isn't as clear as I thought it'd be. extract to a method called `stopPosting()` instead?

src/metrics.ts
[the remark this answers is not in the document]
chriswk: For 404:
- if your api is down, you should be getting 500 (service error) or a type of 502 or 504 (gateway timeouts), not 404, at least from our hosted instances. I guess there are reverse proxies out there that returns 404 if the service they're proxying is not available. 

For 401/403:
Currently, you can't give tokens access after creation, so I'd be very surprised if you went from 403 to 200. You either have a client token or not. A client token which gets new access would simply see more features returned, not suddenly move from a FORBIDDEN to an OK response.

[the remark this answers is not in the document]
chriswk: But doing what the Java SDK did instead, immediately log at error level that the API most likely is incorrectly configured and then back off to our maximum backoff (10 * interval), could also be an answer

chriswk:
> Looks good! I noticed the implementations are a little bit different between the metrics and the features pollings. Is that intentional?

Yes, the metrics is a POST action, whereas the polling is a GET. We're also more OK with losing metrics than not being able to get latest feature configuration.

src/metrics.ts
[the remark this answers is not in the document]
chriswk: hmm, we're calling an explicit `this.stop()` here already, so dropping the -1 seems fair. Will fix.

src/metrics.ts
```diff
@@ -170,6 +190,19 @@ export default class Metrics extends EventEmitter {
...
+    this.emit(UnleashEvents.Warn, `${url} returning ${statusCode}, stopping metrics`);
+    this.failures = -1;
+    this.stop();
+  }
+
+  backoff(url: string, statusCode: number): void {
+    this.failures = Math.min(10, this.failures + 1);
+    // eslint-disable-next-line max-len
+    this.emit(UnleashEvents.Warn, `${url} returning ${statusCode}. Backing off to ${this.getInterval()}`);
+    this.startTimer();
```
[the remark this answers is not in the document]
chriswk: ah, yes, we call a random function each time. Let's remove the call to getInterval here, and just use failure count to say we're backing off to `x times interval`, no need to be exact to the ms anyway.

src/metrics.ts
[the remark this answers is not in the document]
gastonfournier: Also, for metrics is ok to be more aggressive and stop posting

gastonfournier (approved):
LGTM!

nunogois (approved):
I noticed some approvals got stale after recent changes, so here's my approval in case it unblocks you 👍

src/metrics.ts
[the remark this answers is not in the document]
chriswk: renamed the method to `configurationError` after discussion with Gaston

PR #539, opened 2023-11-27
fix: restore removed public exports

Tymek:
## About the changes
Public interface `VariantWithFeatureStatus` and function `getDefaultVariant` was accidentally refactored out instead of being deprecated.

### DIFF
--- a/src/variant.ts
+++ b/src/variant.ts
@@ -41,6 +41,16 @@ export const defaultVariant: Variant = {
   feature_enabled: false,
 };
 
+/**
+ * @deprecated Use `Variant` directly
+ */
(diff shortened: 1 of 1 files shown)


src/variant.ts
```diff
@@ -41,6 +41,16 @@ export const defaultVariant: Variant = {
   feature_enabled: false,
 };
 
+/**
+ * @deprecated Use `Variant` directly
+ */
+export interface VariantWithFeatureStatus extends Variant {}
```
Tymek, line 47: Previously it was
```
export interface VariantWithFeatureStatus extends Variant {
  featureEnabled: boolean;
}

So it's tricky to bring it back.

PR #570, opened 2024-01-23
fix: variant stickiness from strategy

kwasniew:
## About the changes

If variant doesn't have stickiness fall back to strategy stickiness.

Fixes updated and failing spec: https://github.com/Unleash/client-specification/pull/83


FredrikOseberg (approved):
LGTM

PR #574, opened 2024-01-30
feat: give option to retrieve toggle definitions with full segment data

andreas-unleash:
This will give the option to retrieve toggle definitions with full segment data

Relates to [SR-353](https://linear.app/unleash/issue/SR-353/expose-getsegment-method-request-for-access-to-raw-segment-data)

### DIFF
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from 'events';
-import { ClientFeaturesResponse, FeatureInterface } from '../feature';
+import { ClientFeaturesResponse, EnhancedFeatureInterface, FeatureInterface } from '../feature';
 import { get } from '../request';
 import { CustomHeaders, CustomHeadersFunction } from '../headers';
 import getUrl from '../url-utils';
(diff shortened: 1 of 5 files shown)


src/repository/index.ts
```diff
@@ -405,4 +406,18 @@ Message: ${err.message}`,
   getToggles(): FeatureInterface[] {
     return Object.keys(this.data).map((key) => this.data[key]);
   }
+
+  enhanceWithSegmentData(toggles: FeatureInterface[]): EnhancedFeatureInterface[] {
```
sighphyre, line 410: There's a lot going on in this method. I think this deserves some unit tests

src/unleash.ts
```diff
@@ -332,8 +332,15 @@ export class Unleash extends EventEmitter {
     return this.repository.getToggle(toggleName);
   }
 
-  getFeatureToggleDefinitions(): FeatureInterface[] {
-    return this.repository.getToggles();
+  getFeatureToggleDefinitions(
+    withFullSegments = false
```
sighphyre, line 336: This is quite clever from a perspective of not having to muck with the version number but I'm tempted to bite the bullet and go with this signature instead:

`getFeatureToggleDefinitions(): EnhancedFeatureInterface[]`

It would definitely mean a major version but it'd make the method a lot easier to use. WDYT?

sighphyre (commented):
Yeah this looks good but I don't want to mark approve without some tests. This is definitely stuff we'll forget and we likely won't use it ourselves, I'd be worried about accidentally breaking this without noticing in the future

src/unleash.ts
andreas-unleash: The reason of doing it this way is that segment data can be quite massive (seeing how many userIds some customers have as values.
This is meant to solve a use case but maybe overloading the data should be optional. wdyt?

sighphyre: Actually hadn't thought of that. Yeah fair enough, that's potentially a very, very expensive nested loop so that would be a surprise to anyone who didn't care about this stuff already. Fair enough!

sighphyre (approved):
G

PR #580, opened 2024-02-06
feat: prevent double start

kwasniew:
## About the changes

Prevent unleash client from starting twice. It will prevent bugs that were previously fixed in the user space: https://github.com/Unleash/unleash/pull/6141

Details:
* added boolean status similar to existing statuses not to introduce a new pattern in code
* the new started status is only used internally to prevent double repository and metrics start

### DIFF
--- a/src/unleash.ts
+++ b/src/unleash.ts
@@ -47,6 +47,8 @@ export class Unleash extends EventEmitter {
 
   private ready: boolean = false;
 
+  private started: boolean = false;
+
   constructor({
     appName,
     environment = 'default',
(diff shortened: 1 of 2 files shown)


src/unleash.ts
```diff
@@ -245,6 +247,8 @@ export class Unleash extends EventEmitter {
   }
 
   async start(): Promise<void> {
+    if(this.started) return;
+    this.started = true;
```
ivarconr, line 251: Oh, so this was the error. Ok, this is a great improvement. 

Did you also check the destroy() method?

ivarconr (approved):

src/unleash.ts
kwasniew: Do we want to start the client again after destroy? If not then destroy changes are not needed

PR #594, opened 2024-02-22
fix(lint): run lint-staged on pre-commit again

theneva:
Hi! :wave:

This PR is based on #593, so please ignore the first commit.

Husky dropped support for the `hooks` field in `package.json` in v5, so the pre-commit hook (and, by extension `lint-staged`) has not been run since Husky was upgraded from v4 to v7 in 2021 (https://github.com/unleash/unleash-client-node/commit/fcf3c189bd3bde945f01b4549269025126400055).

I also made ESLint actually run Prettier by extending the `prettier/recommended` config.


.eslintrc
```diff
@@ -1,11 +1,9 @@
 {
   "extends": [
-    "airbnb-typescript/base",
-    "prettier",
     "plugin:import/recommended",
-    "plugin:import/typescript"
+    "plugin:import/typescript",
+    "plugin:prettier/recommended"
```
theneva, line 5: From https://github.com/prettier/eslint-plugin-prettier#configuration-legacy-eslintrc:

> This will:
> 
> - Enable the prettier/prettier rule.
> - Disable the arrow-body-style and prefer-arrow-callback rules which are problematic with this plugin - see the below for why.
> - Enable the eslint-config-prettier config which will turn off ESLint rules that conflict with Prettier.

package.json
```diff
@@ -116,18 +115,13 @@
     ]
   },
   "lint-staged": {
-    "*.js": [
+    "*.{js,ts}": [
```
theneva, line 118: ESLint didn't support TypeScript when this was first added in 2018, but it does now!

package.json
```diff
@@ -116,18 +115,13 @@
...
-    "*.{ts,md,json}": [
+    "*.{md,json}": [
       "prettier --write"
     ]
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
-  },
```
theneva, line 130: this configuration style was dropped in Husky v5, which is why it stopped working. The hooks are now configured as separate files in `.husky/`

.husky/pre-commit
```diff
@@ -0,0 +1 @@
+yarn run pre-commit
```
theneva, line 1: This is the configuration style Husky has used since v5, so this file replaces the `hooks.pre-commit` property in `package.json`

package.json
```diff
@@ -9,9 +9,10 @@
     "version": "./scripts/update-build-details.sh $npm_package_version",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix  && prettier --write '{src,examples}/**/*.ts'",
+    "pre-commit": "lint-staged",
     "prebuild": "del-cli --force lib",
     "build": "tsc -p .",
-    "prepare": "yarn run build",
+    "prepare": "husky && yarn run build",
```
theneva, line 15: Running `husky` installs the hooks in `.husky/…` into `.git/hooks/…`

package.json
```diff
@@ -62,14 +63,12 @@
...
     "eslint": "^8.44.0",
-    "eslint-config-airbnb-base": "^15.0.0",
-    "eslint-config-airbnb-typescript": "^17.0.0",
     "eslint-config-prettier": "^8.8.0",
     "eslint-import-resolver-typescript": "^3.5.5",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-prettier": "^5.0.0",
     "esm": "^3.2.25",
-    "husky": "^8.0.3",
+    "husky": "^9.0.11",
```
theneva, line 71: I figured I'd bump Husky to the latest v9 since I was touching it anyway – but the changes in this PR should work fine with v8 too

.eslintrc
```diff
@@ -1,11 +1,9 @@
 {
   "extends": [
-    "airbnb-typescript/base",
```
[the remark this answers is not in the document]
SimenB: you probably need the `parser` etc? line 10 and down https://www.npmjs.com/package/eslint-config-airbnb-typescript?activeTab=code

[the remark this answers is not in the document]
theneva: you're absolutely right. I tested locally with a file with no typescript-specific keywords… I just added eslint-config-airbnb-typescript back in now

gastonfournier (approved):
This looks great @theneva! In Unleash we switched to biome which is faster, but we believe this is great anyway and don't have a strong attachment to biome. I'd say we go ahead with this one! Thanks for the contribution!

PR #604, opened 2024-04-15
Making it clearer that the API token needs to be passed via custom headers

alvinometric:
In the first example that people will see, the `Authorization` header is like this 
```
customHeaders: { Authorization: 'SOME-SECRET' },
```

To me it's not super clear that API token **needs** to be passed via custom headers.

I've also standardised that value throughout the readme.


andreas-unleash (approved):
Thanks for clarifying.

PR #605, opened 2024-04-16
Updated README to use ESM

alvinometric:
Since we're only supporting Node.js 16 or newer, I think ESM makes sense. But happy to wait a bit too


FredrikOseberg (approved):
Updating docs to use ESM syntax is sensible to me. Nice!

PR #616, opened 2024-05-22
Fixes Unleash client stops fetching #615

### LINKED ISSUE #615: Unleash client stops fetching
### Describe the bug

We encountered an issue with Unleash Proxy a few times when it seemed that it stopped getting any updates from Unleash. It was caused by the update in unleash-client release 5.2 because it stops fetching Unleash configuration when it encounters status codes 401, 403 or 404. Please see `handleErrorCases` function. I agree that it seems like a configuration issue, but in my opinion, after the successful initial fetch, it doesn't make sense to just stop pulling updates from Unleash in case a single 404 is encountered.

_No response_

### Expected behavior

_No response_

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

_No response_

### Subscription type

None

### Hosting type

None

_No response_

HenrikBacher:
Fixes #615 by treating http 404 as a recoverable error

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -225,9 +225,10 @@ export default class Metrics extends EventEmitter {
         httpOptions: this.httpOptions,
       });
       if (!res.ok) {
-        if (res.status === 404 || res.status === 403 || res.status == 401) {
+        if (res.status === 403 || res.status == 401) {
           this.configurationError(url, res.status);
         } else if (
+          res.status === 404 ||
           res.status === 429 ||
           res.status === 500 ||
           res.status === 502 ||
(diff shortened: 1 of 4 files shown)


chriswk (approved):
Thanks for this PR. Kinda worried about your server setup if you get intermittent 404s for a server instance, but I'll merge this since you added it to the backoff logic. At least we won't see the client get a 404 and therefore end up in a retry loop we did earlier.

chriswk:
Waiting for the build to go green before merging.

chriswk (approved):
Yes

chriswk:
The reason for the fail here was reported on client-specifications. Will bump to 5.1.6 on main, but will merge this.

PR #617, opened 2024-05-23
Fix wrong error when receiving 404 after #616

HenrikBacher:
## About the changes
After my fix in #616 the same error was return from http 429 & 404, added specific error for http 404


chriswk (approved):
Thanks, good catch. Thank you for cleaning up :)

PR #627, opened 2024-06-13
Chore/version upgrades

chriswk:
In order to come more in line with unleash-server, upgrades make-fetch-happen to 13.0.1 as well as bump braces to the 3.0.3 and typescript to 5.4.3


nunogois (approved):
LG

PR #629, opened 2024-06-17
fix: handle falsy context fields

nunogois:
Follow up to https://github.com/Unleash/unleash-client-node/pull/628

I noticed we were not correctly handling falsy values in our context values due to doing a truthy check on the value when resolving it. 

By replacing that check with a more explicit `!== undefined` check we can still properly evaluate context fields such as `false` or `0`. I believe this check tracks with the original intent of simply checking whether the field is present.

Eventually all of this could be simplified to: `return context[field]?.toString() ?? context.properties?.[field]?.toString();`, but I wanted to avoid that level of refactor in this PR. I can still do it if the reviewer agrees.

Also added new tests that cover these specific cases.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -19,10 +19,10 @@ export function createFallbackFunction(
 }
 
 export function resolveContextValue(context: Context, field: string): string | undefined {
-  if (context[field]) {
+  if (context[field] !== undefined) {
     return context[field]?.toString();
   }
-  if (context.properties && context.properties[field]) {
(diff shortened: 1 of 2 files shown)


chriswk:
I like your suggestion of collapsing down to the optional operators. :+1:

nunogois:
> I like your suggestion of collapsing down to the optional operators. 👍

Addressed on a separate PR: https://github.com/Unleash/unleash-client-node/pull/630

PR #630, opened 2024-06-17
refactor: resolveContextValue

nunogois:
Follow up to https://github.com/Unleash/unleash-client-node/pull/629

Refactors `resolveContextValue` to: `return context[field]?.toString() ?? context.properties?.[field]?.toString();`, as discussed on that PR.


gastonfournier (approved):
LG!

PR #637, opened 2024-07-09
feat: automatically generated instanceId

nunogois:
https://linear.app/unleash/issue/2-2397/make-instance-id-autogenned-node

This makes `instanceId` an automatically generated property, instead of being set by the user.

Similar to https://github.com/Unleash/unleash-client-ruby/pull/179
Similar to https://github.com/Unleash/unleash-client-dotnet/pull/226

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,6 @@ The initialize method takes the following arguments:
 - **environment** - The value to put in the Unleash context's `environment` property. Automatically
   populated in the Unleash Context (optional). This does **not** set the SDK's
   [Unleash environment](https://docs.getunleash.io/reference/environments).
-- **instanceId** - A unique identifier, should/could be somewhat unique.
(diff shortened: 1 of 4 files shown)


alvinometric:
Not sure it needs a major version because the code will still work if you pass an object with an `instanceId`

sighphyre (approved):
Ye!

sighphyre:
> Not sure it needs a major version because the code will still work if you pass an object with an `instanceId`

It matters for TypeScript, sadly (actually strongly mixed feelings?). I thought the same thing so I went and checked and yes the compiler does complain

alvinometric:
:old-man-yells-at-typescript:

naderghanbari:
What about Unleash-API-compatible providers like Gitlab Feature Flags, that do require an `instanceId`? Does this mean the sdks/clients will be incompatible with these providers after this change?

PR #638, opened 2024-07-10
feat: http and https proxy support

kwasniew:
## About the changes

Support for HTTP_PROXY and HTTPS_PROXY

E.g. If you have a proxy at http://localhost:8081 pointing to your Unleash API URL run your app that contains the Node SDK with: `HTTP_PROXY=http://localhost:8081 node my_app.js`

Design decision:
* instead of using @npmcli/agent directly that does the env var detection I decided to implement detection myself and instantiate http and https proxies directly. Those are the same proxies that the @npmcli/agent instantiates. The reason I choose to do it myself is because @npmcli/agent doesn't have any TS types and we loose autocompletion. On the other hand the low level agents have built-in TS support. 

Can be tested locally with https://github.com/kwasniew/proxy-server

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,11 @@ offline, from a browser environment or implement your own caching layer. See
 Unleash depends on a `ready` event of the repository you pass in. Be sure that you emit the event
 **after** you've initialized unleash.
 
+## Usage with HTTP and HTTPS proxies
+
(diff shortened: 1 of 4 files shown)


chriswk (approved):
Nice. Looks good

PR #639, opened 2024-07-10
feat: extend metrics with metadata

nunogois:
https://linear.app/unleash/issue/2-2389/extend-node-sdk-with-new-metrics

This extends the current metrics with metadata.

Tries being a bit fancy and grab other runtimes like `bun` and `deno`. E.g. if we run something in Bun that uses our SDK we can see it in the metrics payload:
image

Similar to: https://github.com/Unleash/unleash-client-dotnet/pull/231 and https://github.com/Unleash/unleash-client-ruby/pull/177

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ The initialize method takes the following arguments:
 
 ### instanceId
 
-As of version 5.0.0, `instanceId` is now automatically generated.
+As of version 6.0.0, `instanceId` is now automatically generated.
 
 ## Custom strategies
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -231,7 +231,7 @@ The initialize method takes the following arguments:
 
 ### instanceId
 
-As of version 5.0.0, `instanceId` is now automatically generated.
```
nunogois, line 234: Scouting: Noticed this was wrong previously. We're already in v5, so next major is 6.

sighphyre (approved):
LGTM

PR #648, opened 2024-07-23
fix: edge case on context value where trim is not a function

nunogois:
https://linear.app/unleash/issue/2-2468/fix-trim-is-not-a-function-edge-case-on-context-value

Fixes: https://github.com/Unleash/unleash-proxy/issues/185

This PR addresses the issue reported by @nicks, where a `version.trim is not a function` error is thrown under certain conditions.

The error occurs when a context value has a custom `.toString()` method that does not return a string. This was the only scenario in which we could reproduce the issue. A corresponding test has been added to validate this.

The solution suggested by @nicks has been implemented. Specifically, the `!isStrictSemver(contextValue)` check has been moved inside the try/catch block. This change ensures that any exceptions thrown by this method will be caught, in which case the function will default to false.

A more robust implementation of `resolveContextValue` has been introduced, which would also prevent the issue. See: 

image

Lastly, we noticed the test files `strategy-test.ts` and `flexible-rollout-test.ts` were being ignored in test runs due to their filename format. Renaming them to `strategy.test.ts` and `flexible-rollout.test.ts` ensures they are included in the tests as intended.


chriswk (approved):
Nice. Strange corner case.

PR #661, opened 2024-09-23
Dependency update

alvinometric:
- Updated to Yarn V4, just like the rest of Unleash
- Removed `esm` as a dependency, I couldn't find it anywhere in the codebase


alvinometric:
bump

PR #671, opened 2024-10-28
chore: drop strict check in semver in favor of checking it is valid

nunogois:
This drops the strictness check in our semver strategy constraints in favor of checking whether the semver is valid.

We found this out by trying to create a semver constraint for ourselves using the Unleash semver format (x.x.x+hash)


gastonfournier (approved):
LGTM! Go!

PR #679, opened 2024-12-02
feat: streaming poc

kwasniew:
## About the changes

This is a reference implementation for the streaming client using SSE (Server-Sent Events).

The changes include:
* production code changes in the repository
* example how to test the streaming API (we test the API, not the low-level details that are tested in the ES library itself)
* example showing user facing property to enable streaming

Chosen library:
* we choose a fork of https://github.com/EventSource/eventsource at https://github.com/launchdarkly/js-eventsource
because it provides extra options such as backoff with jitter, read timeout for inactive streams. Both libraries also support HTTP headers that are not part of the SSE spec itself but useful in practice to do auth

Main logic:
* customers can specify mode (polling vs streaming) using experimentalMode property. When they opt into streaming we create EventSource and inject it to the repository
* in the repository we always make the initial fetch with the HTTP fetch API same as before
* however when streaming is enabled (= EventSource is injected to the repository) we never do setTimeout calls for the next fetch and we listen to event called 'unleash-updated' from the streaming API and update flags accordingly
* so effectively our event stream is a flag stream and not just a notification stream that requires the client to fetch data on its own. This way we save a network roundtrip and piggyback on the broadcast message data

More comments with the important decisions inline

### DIFF
--- a/examples/streaming.js
+++ b/examples/streaming.js
@@ -1,0 +1,14 @@
+const { Unleash } = require('../lib');
+
+const client = new Unleash({
+  appName: 'my-application',
+  url: 'https://app.unleash-hosted.com/demo/api/',
+  customHeaders: {
+    Authorization: '943ca9171e2c884c545c5d82417a655fb77cec970cc3b78a8ff87f4406b495d0',
+  },
+  experimentalMode: { type: 'streaming' },
+  skipInstanceCountWarning: true,
+});
+client.on('changed', () => {
+  console.log(client.isEnabled('demo001', { userId: `${Math.random()}` }));
+});
--- a/examples/variant.js
+++ b/examples/variant.js
@@ -1,14 +1,18 @@
-const { Unleash } = require('../lib');
+const { initialize, getVariant } = require('../lib');
 
-const client = new Unleash({
+const client = initialize({
   appName: 'my-application',
-  url: 'https://sandbox.getunleash.io/nuno/api/',
+  url: 'https://unleash.herokuapp.com/api/',
   customHeaders: {
-    Authorization: '',
+    Authorization: '3bd74da5b341d868443134377ba5d802ea1e6fa2d2a948276ade1f092bec8d92',
   },
-  experimentalMode: { type: 'streaming' },
-  skipInstanceCountWarning: true,
 });
-client.on('changed', () => {
-  console.log('Enabled:', client.isEnabled('streaming-flag', { userId: `${Math.random()}` }));
-});
+
+client.on('error', console.error);
+client.on('warn', console.log);
+
+console.log('Fetching toggles from: http://unleash.herokuapp.com');
+
+setInterval(() => {
+  console.log('Variant config', getVariant('Test.variants', { userId: `${Math.random()}` }));
+}, 1000);
--- a/examples/variant.js
+++ b/examples/variant.js
@@ -1,19 +1,14 @@
-const { initialize, isEnabled } = require('../lib');
+const { Unleash } = require('../lib');
 
-const client = initialize({
+const client = new Unleash({
   appName: 'my-application',
-  url: 'http://localhost:4242/api/',
+  url: 'https://sandbox.getunleash.io/nuno/api/',
   customHeaders: {
-    Authorization: '*:development.35a4fe08c112c8d98cc7a21bdf4d077796920c5e86b0f98eed467b23',
+    Authorization: '',
   },
-  streaming: true,
+  experimentalMode: { type: 'streaming' },
+  skipInstanceCountWarning: true,
 });
-
-client.on('error', console.error);
-client.on('warn', console.log);
-
-console.log('Fetching toggles from: http://unleash.herokuapp.com');
-
-setInterval(() => {
-  console.log('Enabled:', isEnabled('sadfsdaf', { userId: `${Math.random()}` }));
-}, 1000);
+client.on('changed', () => {
+  console.log('Enabled:', client.isEnabled('streaming-flag', { userId: `${Math.random()}` }));
+});
--- a/examples/variant.js
+++ b/examples/variant.js
@@ -1,11 +1,12 @@
-const { initialize, getVariant } = require('../lib');
+const { initialize, isEnabled } = require('../lib');
 
 const client = initialize({
   appName: 'my-application',
-  url: 'https://unleash.herokuapp.com/api/',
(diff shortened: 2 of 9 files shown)


examples/variant.js
```diff
@@ -1,11 +1,12 @@
...
 
 const client = initialize({
   appName: 'my-application',
-  url: 'https://unleash.herokuapp.com/api/',
+  url: 'http://localhost:4242/api/',
   customHeaders: {
-    Authorization: '3bd74da5b341d868443134377ba5d802ea1e6fa2d2a948276ade1f092bec8d92',
+    Authorization: '*:development.35a4fe08c112c8d98cc7a21bdf4d077796920c5e86b0f98eed467b23',
   },
+  streaming: true,
```
kwasniew, line 9: opt into streaming

src/repository/index.ts
```diff
@@ -13,6 +13,7 @@ import {
   Segment,
   StrategyTransportInterface,
 } from '../strategy/strategy';
+const EventSource = require('eventsource');
```
kwasniew, line 16: couldn't get it to work with import, only require worked in a quick PoC

src/repository/index.ts
```diff
@@ -122,10 +129,23 @@ export default class Repository extends EventEmitter implements EventEmitter {
     this.bootstrapOverride = bootstrapOverride;
     this.storageProvider = storageProvider;
     this.segments = new Map();
+    this.streaming = streaming;
+    if (this.streaming) {
+      this.eventSource = new EventSource(resolveUrl(this.url, './client/streaming'), {
+        headers: this.headers,
+      } as any);
+      this.eventSource?.addEventListener('message', (event) => {
```
kwasniew, line 137: required EventSource is seen as potentially undefined. Not sure why that's what ? is handling here

src/repository/index.ts
```diff
@@ -122,10 +129,23 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
+      });
+      this.eventSource?.addEventListener('error', (error) => {
+        console.log('ES error', error);
+      });
+    }
   }
 
   timedFetch(interval: number) {
-    if (interval > 0) {
+    if (interval > 0 && !this.streaming) {
```
kwasniew, line 148: this line is all it takes to stop intervals/timeouts and use ES client instead

examples/streaming.js
```diff
@@ -0,0 +1,14 @@
+const { Unleash } = require('../lib');
+
+const client = new Unleash({
+  appName: 'my-application',
+  url: 'https://app.unleash-hosted.com/demo/api/',
+  customHeaders: {
+    Authorization: '943ca9171e2c884c545c5d82417a655fb77cec970cc3b78a8ff87f4406b495d0',
+  },
+  experimentalMode: { type: 'streaming' },
```
kwasniew, line 9: this has to be provided explicitly to opt into streaming. The reason I model it as {type: 'polling'} | {type: 'streaming'} union is to add streaming config option later {type: 'streaming', config: {refresh, backoff}}

examples/streaming.js
```diff
@@ -0,0 +1,14 @@
...
+  skipInstanceCountWarning: true,
+});
+client.on('changed', () => {
```
kwasniew, line 12: changes are reflected instantly in the event emitter

src/repository/index.ts
```diff
@@ -13,6 +13,8 @@ import {
   Segment,
   StrategyTransportInterface,
 } from '../strategy/strategy';
+// @ts-expect-error
```
kwasniew, line 16: this library does not provide any types so we have to do this for now until we provide a type ourselves

src/repository/index.ts
```diff
@@ -105,6 +110,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
     bootstrapProvider,
     bootstrapOverride = true,
     storageProvider,
+    eventSource,
```
kwasniew, line 113: we make EventSource injectable so that it's easy to swap in testing for a simple mock

src/repository/index.ts
```diff
@@ -122,10 +128,30 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
+            this.etag = etag;
+          } else {
+            this.etag = undefined;
+          }
+          this.save(data, true);
+        } catch (err) {
+          this.emit(UnleashEvents.Error, err);
+        }
+      });
+      this.eventSource.addEventListener('error', (error: unknown) => {
```
kwasniew, line 147: for now I decided to translate event source errors to warnings

src/repository/index.ts
```diff
@@ -122,10 +128,30 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
+        this.emit(UnleashEvents.Warn, error);
+      });
+    }
   }
 
   timedFetch(interval: number) {
-    if (interval > 0) {
+    if (interval > 0 && !this.eventSource) {
```
kwasniew, line 154: do intervals only when eventSource is not provided

src/repository/index.ts
```diff
@@ -398,6 +424,9 @@ Message: ${err.message}`,
       clearTimeout(this.timer);
     }
     this.removeAllListeners();
+    if (this.eventSource) {
```
kwasniew, line 427: we have to remember about closing the event source when Unleash is closed

src/unleash.ts
```diff
@@ -123,6 +126,17 @@ export class Unleash extends EventEmitter {
         tags,
         bootstrapProvider,
         bootstrapOverride,
+        eventSource:
+          experimentalMode?.type === 'streaming'
+            ? new EventSource(resolveUrl(unleashUrl, './client/streaming'), {
```
kwasniew, line 131: we can expose the low level config options to the public API over time, but for now I set those defaults internally

src/test/repository.test.ts
```diff
@@ -1360,3 +1361,73 @@ test('Stopping repository should stop storage provider updates', async (t) => {
...
+    name: 'feature',
+    enabled: true,
+    strategies: [
+      {
+        name: 'default',
+      },
+    ],
+  };
+  const storageProvider: StorageProvider<ClientFeaturesResponse> = new InMemStorageProvider();
+  const eventSource = {
```
kwasniew, line 1378: in-memory mock is trivial to implement. It's just a thin layer on top of the EventEmitter to adjust the API semantic to SSE API

PR #680, opened 2024-12-03
Add support for no_proxy/NO_PROXY environment

xpol:
## About the changes

Add support for `no_proxy`/`NO_PROXY` environment by using the [proxy-from-env](https://github.com/Rob--W/proxy-from-env) library.

### Important files

src/request.ts

## Discussion points

Support `HTTP_PROXY` and `HTTPS_PROXY` in #638  break enterprise users who use both `HTTP(S)_PROXY` and `NO_PROXY`.
We need full support for `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`.

And [proxy-from-env](https://github.com/Rob--W/proxy-from-env) is battle tested by the [axios library](https://github.com/axios/axios/blob/v1.x/lib/adapters/http.js#L7).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10045 @@
+{
+  "name": "unleash-client",
+  "version": "6.1.3",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "unleash-client",
+      "version": "6.1.3",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "http-proxy-agent": "^7.0.2",
+        "https-proxy-agent": "^7.0.5",
+        "ip-address": "^9.0.5",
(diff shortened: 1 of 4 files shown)


kwasniew:
Thanks for a great contribution. I cleaned up some lockfiles in another PR https://github.com/Unleash/unleash-client-node/pull/682 and merged it with you as a co-author. We'll release a new version soon. How do you use Unleash Node SDK? Directly or as part of the proxy (https://github.com/Unleash/unleash-proxy)? I'm asking because this env var proxy capability was originally added for the proxy but proxy is in maintenance mode now and we update it only when needed.

kwasniew:
Your code has been published in 6.2.0. Please let me know if it works for you.

PR #687, opened 2024-12-19
fix: retry all streaming errors

kwasniew:
## About the changes

Improving retry on streaming behavior.
Previously we were only retrying on 500 codes.

Details:
* handling unleash-connected event but we ignore it on first connect. First connect is done with a regular fetch for a safety fallback e.g. when server doesn't support streaming
* telling eventsource client to retry all types of errors so when some network proxy ends the connection without sending HTTP code we still re-connect and refetch the features from the unleash-connected event data (no data updated during proxy downtime is lost as a result)

### DIFF
--- a/src/test/repository.test.ts
+++ b/src/test/repository.test.ts
@@ -1362,7 +1362,7 @@ test('Stopping repository should stop storage provider updates', async (t) => {
   t.is(result, undefined);
 });
 
-test('Streaming update events', async (t) => {
+test('Streaming', async (t) => {
   t.plan(7);
   const url = 'http://unleash-test-streaming.app';
   const feature = {
--- a/src/test/repository.test.ts
+++ b/src/test/repository.test.ts
@@ -1362,9 +1362,9 @@ test('Stopping repository should stop storage provider updates', async (t) => {
   t.is(result, undefined);
 });
 
-test('Streaming', async (t) => {
-  t.plan(6);
-  const url = 'irrelevant';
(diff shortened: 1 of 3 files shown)


src/test/repository.test.ts
```diff
@@ -1402,16 +1403,24 @@ test('Streaming', async (t) => {
...
+
+  // first connection is ignored, since we do regular fetch
+  eventSource.emit('unleash-connected', {
+    type: 'unleash-connected',
+    data: JSON.stringify({ meta: {}, features: [{ ...feature, name: 'intialConnectedIgnored' }] }),
+  });
+
   const before = repo.getToggles();
-  t.deepEqual(before, []);
+  t.deepEqual(before, [{ ...feature, name: 'initialFetch' }]);
```
kwasniew, line 1415: test showing initial fetch is outside of the streaming flow

src/test/repository.test.ts
```diff
@@ -1402,16 +1403,24 @@ test('Streaming', async (t) => {
     eventSource,
   });
 
+  await repo.start();
+
+  // first connection is ignored, since we do regular fetch
+  eventSource.emit('unleash-connected', {
+    type: 'unleash-connected',
+    data: JSON.stringify({ meta: {}, features: [{ ...feature, name: 'intialConnectedIgnored' }] }),
```
kwasniew, line 1411: show that initial streaming connect is ignored

thomasheartman (approved):

PR #690, opened 2025-01-07
feat: client identification headers

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -7,6 +7,7 @@ import { URL } from 'url';
 import { getProxyForUrl } from 'proxy-from-env';
 import { CustomHeaders } from './headers';
 import { HttpOptions } from './http-options';
+const details = require('./details.json');
 
 export interface RequestOptions {
   url: string;
@@ -18,6 +19,7 @@ export interface GetRequestOptions extends RequestOptions {
   etag?: string;
   appName?: string;
   instanceId?: string;
+  connectionId: string;
   supportedSpecVersion?: string;
   httpOptions?: HttpOptions;
 }
@@ -30,6 +32,7 @@ export interface PostRequestOptions extends RequestOptions {
   json: Data;
   appName?: string;
   instanceId?: string;
+  connectionId?: string;
   httpOptions?: HttpOptions;
 }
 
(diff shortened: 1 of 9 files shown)


src/test/unleash.test.ts
```diff
@@ -51,7 +51,10 @@ const defaultToggles = [
 ];
 
 function mockNetwork(toggles = defaultToggles, url = getUrl()) {
-  nock(url).get('/client/features').reply(200, { features: toggles });
+  nock(url)
+    .get('/client/features')
+    .matchHeader('x-unleash-connection-id', /^.{36}$/)
```
kwasniew, line 56: every request to unleash should have 36 digit uuid

src/request.ts
```diff
@@ -55,18 +58,35 @@ export const getDefaultAgent = (url: URL) => {
...
+  contentType,
+  custom,
+  specVersionSupported,
+  connectionId,
+}: HeaderOptions): Record<string, string> => {
   const head: Record<string, string> = {};
   if (appName) {
+    // TODO: delete
     head['UNLEASH-APPNAME'] = appName;
+    // TODO: delete
```
thomasheartman, line 84: delete when, though? Next major?

src/test/repository.test.ts
```diff
@@ -177,12 +183,15 @@ test('should request with etag', (t) =>
     repo.start();
   }));
 
-test('should request with custom headers', (t) =>
+test('should request with correct headers', (t) =>
```
thomasheartman, line 186: I see why you're co-opting the tests here, but we're also losing a bit of information about the tests with this rename. Before, it was clear that we're talking about custom headers, so if you're scanning through the tests, it was easy enough to find (or grep for). "Correct headers" is more ... ambiguous. How about renaming this one to something like this?
```suggestion
test('should request with custom and `x-unleash` headers', (t) =>
```

The same goes for the other tests that were changed from "custom" to "correct".

thomasheartman (approved):
Nice one!

src/request.ts
kwasniew: I want to have a super safe version upgrade that I can recommend to everyone even if for some unknown reason they relied on he old headers

src/test/repository.test.ts
kwasniew: Good call. I also updated one more test with the same change

thomasheartman (approved):

PR #696, opened 2025-01-16
fix: cross platform uuid

kwasniew:
## About the changes

UUID generation with crypto has better uniqueness guarantees, but Math.random() solution is good enough for our use case (which is not security critical) and works better across all platforms.

This PR will fix the NextJS SDK build issue.

We use UUID for estimated connection counting which is using imprecise HyperLogLog anyway.

### DIFF
--- a/src/uuidv4.ts
+++ b/src/uuidv4.ts
@@ -1,7 +1,7 @@
 /**
  * This function generates a UUID using Math.random().
  * The distribution of unique values is not guaranteed to be as robust
- * as with a crupto module but works across all platforms (Node, React Native, browser JS).
+ * as with a crypto module but works across all platforms (Node, React Native, browser JS).
  *
  * We use it for connection id generation which is not critical for security.
  */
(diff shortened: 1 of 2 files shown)


src/uuidv4.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * This function generates a UUID using Math.random().
+ * The distribution of unique values is not guaranteed to be as robust
+ * as with a crupto module but works across all platforms (Node, React Native, browser JS).
+ *
+ * We use it for connection id generation which is not critical for security.
+ */
+export const uuidv4 = (): string => {
```
kwasniew, line 8: existing tests check uuid is 36 characters

thomasheartman (dismissed):
Yeah, I'm tempted to say that this is fine. Given our use case, I don't expect we'll run into issues, and if we *do* find that we're getting dupes, then we can always change ti later.

thomasheartman (approved):

PR #706, opened 2025-02-03
fix: drop ts type inference for event source

Tymek:
## About the changes
- fix error with dependency typings
- add test to catch install issues in the future
- update Node types, to build without errors

Closes https://github.com/Unleash/unleash-client-node/issues/705
Internal ticket: [issue/1-3339/node-sdk-missing-dependency-for-types](https://linear.app/unleash/issue/1-3339/node-sdk-missing-dependency-for-types)

### DIFF
--- a/scripts/test-package/run.sh
+++ b/scripts/test-package/run.sh
@@ -2,8 +2,8 @@
 set -e
 
 TEST_DIR="test-package"
-
 mkdir "$TEST_DIR"
+
 cp scripts/test-package/test-tsconfig.json "$TEST_DIR/tsconfig.json"
 cp scripts/test-package/test-package.json "$TEST_DIR/package.json"
 cd "$TEST_DIR"
@@ -11,5 +11,13 @@ npm install --install-links
 mkdir src
 echo -e "import { Unleash } from 'unleash-client';\nvoid Unleash;\nconsole.log('Hello world');" > src/index.ts
 ./node_modules/.bin/tsc -b tsconfig.json
-cd ..
-rm -rf "$TEST_DIR"
+
+if [ "$(node . 2>&1)" = "Hello world" ]; then
+  echo "Output is correct"
+  (cd .. && rm -rf "$TEST_DIR")
+else
+  echo "Output is incorrect" >&2
+  echo $(node . 2>&1)
+  (cd .. && rm -rf "$TEST_DIR")
+  exit 1
+fi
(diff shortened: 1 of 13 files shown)


src/metrics.ts
```diff
@@ -97,7 +97,7 @@ export default class Metrics extends EventEmitter {
 
   private url: string;
 
-  private timer: NodeJS.Timer | undefined;
```
kwasniew, line 100: how did you find that this needs to change?

Tymek: I had to update Node.js types to fix other error
```
../node_modules/@types/node/module.d.ts:121:13 - error TS2386: Overload signatures must all be optional or required.

121             resolve?(specified: string, parent?: string | URL): Promise<string>;
                ~~~~~~~
```
And new types where complaining about NodeJS.Timer

scripts/test-package/run.sh
```diff
@@ -0,0 +1,23 @@
...
+
+cp scripts/test-package/test-tsconfig.json "$TEST_DIR/tsconfig.json"
+cp scripts/test-package/test-package.json "$TEST_DIR/package.json"
+cd "$TEST_DIR"
+npm install --install-links
+mkdir src
+echo -e "import { Unleash } from 'unleash-client';\nvoid Unleash;\nconsole.log('Hello world');" > src/index.ts
+./node_modules/.bin/tsc -b tsconfig.json
+
+if [ "$(node . 2>&1)" = "Hello world" ]; then
```
kwasniew, line 15: why do we need those checks? Isn't checking exit code from tsc to be 0 enough?

src/metrics.ts
Tymek: BTW, we should add Node v22 to our test array.
https://nodejs.org/en/about/previous-releases

yarn.lock
```diff
@@ -5105,6 +5107,11 @@ undici-types@~5.26.4:
   resolved "https://registry.yarnpkg.com/undici-types/-/undici-types-5.26.5.tgz#bcd539893d00b56e964fd2657a4866b221a65617"
   integrity sha512-JlCMO+ehdEIKqlFxk6IfVoAUVmgz7cU7zD/h9XZ0qzeosSHmUJVOzSQvvYSYWXkFXC+IfLKSIffhv0sVZup6pA==
 
+undici-types@~6.19.2:
```
kwasniew, line 5110: is it from the upgraded `@types/node` transitive deps?

Tymek: yes

kwasniew (approved):
IMO, everything in a test-package required a README explaining why it exists, what problem is solves and what we're checking (since we can't add inline comments in JSON files).

PR #712, opened 2025-02-12
feat: streaming and polling deltas

kwasniew:
## About the changes

Delta support for streaming and polling.

Our Mode type shows what options are possible `Mode = { type: 'polling'; format: 'delta' | 'full' } | { type: 'streaming' };`. In streaming you have no choice and deltas is your only options. In polling mode you can choose and we default to full format.

### DIFF
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -274,7 +274,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
     });
 
     this.setReady();
-    this.emit(UnleashEvents.Changed, Object.values(this.data));
+    this.emit(UnleashEvents.DeltaChanged, delta.events);
     await this.storageProvider.set(this.appName, {
       features: Object.values(this.data),
       segments: [...this.segments.values()],
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -280,7 +280,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
     });
 
     this.setReady();
-    this.emit(UnleashEvents.DeltaChanged, delta.events);
+    this.emit(UnleashEvents.Changed, Object.values(this.data));
     await this.storageProvider.set(this.appName, {
       features: Object.values(this.data),
       segments: [...this.segments.values()],
@@ -439,13 +439,18 @@ Message: ${err.message}`,
       } else if (res.ok) {
         nextFetch = this.countSuccess();
         try {
-          const data: ClientFeaturesResponse = await res.json();
+          const data = await res.json();
           if (res.headers.get('etag') !== null) {
             this.etag = res.headers.get('etag') as string;
           } else {
             this.etag = undefined;
           }
-          await this.save(data, true);
+
+          if (this.mode.type === 'polling' && this.mode.format === 'delta') {
+            await this.saveDelta(data);
+          } else {
+            await this.save(data, true);
+          }
         } catch (err) {
           this.emit(UnleashEvents.Error, err);
         }
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -19,6 +19,7 @@ import {
   StrategyTransportInterface,
 } from '../strategy/strategy';
 import type { EventSource } from '../event-source';
+import { Mode } from '../unleash-config';
 
 export const SUPPORTED_SPEC_VERSION = '5.2.0';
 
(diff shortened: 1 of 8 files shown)


src/repository/index.ts
```diff
@@ -401,13 +439,18 @@ Message: ${err.message}`,
       } else if (res.ok) {
         nextFetch = this.countSuccess();
         try {
-          const data: ClientFeaturesResponse = await res.json();
+          const data = await res.json();
```
kwasniew, line 442: it can be either full response type or delta type

src/repository/index.ts
```diff
@@ -14,8 +19,9 @@ import {
   StrategyTransportInterface,
 } from '../strategy/strategy';
 import type { EventSource } from '../event-source';
+import { Mode } from '../unleash-config';
 
-export const SUPPORTED_SPEC_VERSION = '4.3.0';
+export const SUPPORTED_SPEC_VERSION = '5.2.0';
```
kwasniew, line 24: this hasn't been updated for a while

src/repository/index.ts
```diff
@@ -135,15 +144,11 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
     if (this.eventSource) {
       // On re-connect it guarantees catching up with the latest state.
       this.eventSource.addEventListener('unleash-connected', (event: { data: string }) => {
-        // reconnect
-        if (this.initialEventSourceConnected) {
-          this.handleFlagsFromStream(event);
-        } else {
-          this.initialEventSourceConnected = true;
-        }
+        this.handleFlagsFromStream(event);
```
kwasniew, line 151: in streaming mode we always rely 100% on streaming. previously first request in streaming was a polling request

src/repository/index.ts
```diff
@@ -97,7 +105,7 @@ export default class Repository extends EventEmitter implements EventEmitter {
 
   private eventSource: EventSource | undefined;
 
-  private initialEventSourceConnected: boolean = false;
```
kwasniew, line 100: no need to check if first connection is different from another since all streaming goes with SSEs

src/repository/index.ts
```diff
@@ -154,15 +159,15 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
+      const data: ClientFeaturesDelta = JSON.parse(event.data);
+      this.saveDelta(data);
     } catch (err) {
       this.emit(UnleashEvents.Error, err);
     }
   }
 
   timedFetch(interval: number) {
-    if (interval > 0 && !this.eventSource) {
+    if (interval > 0 && this.mode.type === 'polling') {
```
kwasniew, line 170: we added explicit mode in the constructor params since we need more than eventSource (mode + format)

src/repository/index.ts
```diff
@@ -192,7 +197,11 @@ export default class Repository extends EventEmitter implements EventEmitter {
 
   async start(): Promise<void> {
     // the first fetch is used as a fallback even when streaming is enabled
-    await Promise.all([this.fetch(), this.loadBackup(), this.loadBootstrap()]);
+    await Promise.all([
+      this.mode.type === 'streaming' ? Promise.resolve() : this.fetch(),
```
kwasniew, line 201: no need to fetch data on startup with HTTP call when streaming is enabled

src/repository/index.ts
```diff
@@ -379,7 +417,7 @@ Message: ${err.message}`,
       if (this.tags) {
         mergedTags = this.mergeTagsToStringArray(this.tags);
       }
-      const url = getUrl(this.url, this.projectName, this.namePrefix, mergedTags);
+      const url = getUrl(this.url, this.projectName, this.namePrefix, mergedTags, this.mode);
```
kwasniew, line 420: fetch call need to be conditional

src/url-utils.ts
```diff
@@ -13,8 +15,10 @@ const getUrl = (
   projectName?: string,
   namePrefix?: string,
   tags?: Array<string>,
+  mode?: Mode,
 ): string => {
-  const url = resolveUrl(base, './client/features');
+  const isDeltaPolling = mode && mode.type === 'polling' && mode.format === 'delta';
+  const url = resolveUrl(base, isDeltaPolling ? './client/delta' : './client/features');
```
kwasniew, line 21: delta has to be opt in

sjaanus (approved):
LGTM

PR #715, opened 2025-02-26
feat: interval reporting

kwasniew:
## About the changes

* fetch flag calls include `unleash-interval` header
* send metrics calls include `unleash-interval` header
* `unleash-interval` header is included even when it's set to 0 (but when it's is 0 then no API calls should be made)
* registration calls has `connectionId` added to payload (previously it had `unleash-connection-id` header only)
* registration calls have no `unleash-interval` header since they are invoked once per connection
* send metrics has `connectionId` added to payload (previously it had `unleash-connection-id` header only)


sjaanus (approved):
LG

PR #738, opened 2025-06-11
chore: add a proxy decorator to the client so we can add impact metrics unobtrusively

sighphyre:
This is just a way to wrap the client in a decorator so we can limit exposure of the public API while we experiment on main. This could have been inheritance with less code, just feel this has better separation

Simple usage example moves from this:
``` node
const unleash = initialize({
  appName: 'my-application',
  url,
});
```
to this

``` node
const unleash = new UnleashMetricClient(initialize({
  appName: 'my-application',
  url,
}));

```

### DIFF
--- a/src/metric-client.ts
+++ b/src/metric-client.ts
@@ -1,0 +1,31 @@
+import { Unleash } from './unleash';
+
+function forwardMethods<T extends object>(target: T, source: T, methodNames: (keyof T)[]) {
+  for (const name of methodNames) {
+    // @ts-ignore
+    target[name] = (...args: any[]) => source[name](...args);
+  }
+}
+
+export class UnleashMetricClient {
+  private backingInstance: Unleash;
+
+  constructor(unleash: Unleash) {
+    this.backingInstance = unleash;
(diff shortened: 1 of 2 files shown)


src/metric-client.ts
```diff
@@ -0,0 +1,57 @@
+import { Unleash } from './unleash';
+
+function forwardMethods<T extends object>(target: T, source: T, methodNames: (keyof T)[]) {
```
kwasniew, line 3: you can use proxy instead without having to list all the methods

```
import { Unleash } from './unleash';

export class UnleashMetricClient {
  private proxy: Unleash;

  constructor(unleash: Unleash) {
    this.proxy = new Proxy(unleash, {
      get(target, prop, receiver) {
        const value = Reflect.get(target, prop, receiver);
        // If it's a function, bind it to the target to preserve `this`
        if (typeof value === 'function') {
          return value.bind(target);
        }
        return value;
      }
    });

    return this.proxy as any; // allows consumers to use this like the original Unleash instance
  }
}
```

PR #742, opened 2025-06-17
feat: push impact metrics

sighphyre:
Adds send code for impact metrics.

## Technical Details

- The metrics is now cleared between sends. This is to make the server implementation easier to handle. This does mean that in case of failures we reuse the existing metrics send pattern - we push the data back into the store
- To reuse the metrics send code we need access to the Unleash internals so our decorator has been yeeted


.eslintrc
```diff
@@ -20,6 +20,7 @@
         "ignoreUrls": true
       }
     ],
+    "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
```
sighphyre, line 23: Maybe controversial but I want to implement the interfaces correctly in the tests, but sometimes I legit don't care that the values are consumed, only that there's no escape hatch to bolt in a broken implementation in the public API

src/impact-metrics/metric-client.ts
```diff
@@ -1,57 +1,5 @@
...
-  destroyWithFlush!: Unleash['destroyWithFlush'];
-
-  on!: Unleash['on'];
-
-  start!: Unleash['start'];
-
-  destroy!: Unleash['destroy'];
-
-  isSynchronized!: Unleash['isSynchronized'];
+export class UnleashMetricClient extends Unleash {
```
sighphyre, line 3: To access the metrics sending code we need access to the internals of Unleash, so our decorator plan doesn't work so well anymore

src/metrics.ts
```diff
@@ -369,6 +380,12 @@ export default class Metrics extends EventEmitter {
       yggdrasilVersion: null,
       specVersion: SUPPORTED_SPEC_VERSION,
     };
+
+    if (impactMetrics.length > 0) {
```
sighphyre, line 384: I want to maintain the existing API contract if we don't have metrics to send

src/impact-metrics/metric-types.ts
```diff
@@ -113,7 +117,12 @@ export interface Gauge {
   set(value: number, labels?: MetricLabels): void;
 }
 
-export class ImpactMetricRegistry {
+export interface ImpactMetricRegistry {
```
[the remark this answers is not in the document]
kwasniew: In TS you don't need to do this since you can swap the real registry with anything that has the same structure/shape

src/metrics.ts
```diff
@@ -369,6 +380,12 @@ export default class Metrics extends EventEmitter {
...
+      (base as any).impactMetrics = impactMetrics;
```
kwasniew, line 385: if we're doing type assertion maybe we can assert it's MetricsData & {impactMetrics: ImpactMetricsType}

src/test/metrics.test.ts
```diff
@@ -601,9 +602,124 @@ test('createMetricsData should include extended metrics', (t) => {
...
+    metricsInterval: 0,
+    metricRegistry: fakeMetricRegistry,
+  });
+
+  // Inject a single toggle evaluation so that we force a metrics send but without impact metrics
+  metrics.count('toggle-x', true);
+
+  const scope = nock(url)
+    .post('/client/metrics', (body) => {
+      t.false('impactMetrics' in body);
```
kwasniew, line 718: again I'd move the assertion to the end of the test and only record interaction to a variable here

src/test/metrics.test.ts
```diff
@@ -601,9 +602,124 @@ test('createMetricsData should include extended metrics', (t) => {
...
+  // @ts-expect-error
+  const metrics = new Metrics({
+    url,
+    metricsInterval: 0,
+    metricRegistry: fakeMetricRegistry,
+  });
+
+  const scope = nock(url)
+    .post('/client/metrics', (body) => {
+      t.deepEqual(body.impactMetrics, [impactMetricSample]);
```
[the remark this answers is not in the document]
sighphyre: Sure fair enough, fixed

sighphyre:
> Since we're adding code to the main unleash client maybe it's worth having experimental flag in the config object similar to how we do this in https://github.com/Unleash/unleash-proxy-client-js

Yeah future PR though. Currently there's no real way to access any of this via the public API so people passing new things in doesn't bother me until they can muck something up

src/metrics.ts
```diff
@@ -241,13 +248,15 @@ export default class Metrics extends EventEmitter {
     if (this.disabled) {
       return;
     }
-    if (this.bucketIsEmpty()) {
+    const impactMetrics = this.metricRegistry?.collect() || [];
+
+    if (this.bucketIsEmpty() && impactMetrics.length === 0) {
```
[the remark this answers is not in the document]
sighphyre: Thanks for the spot, yeah it needs to go back in, in this case. Fixed now

src/impact-metrics/metric-types.ts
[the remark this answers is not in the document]
sighphyre: I think I'm going to keep the explicit interface for now. Partially because I want a structured way to apply interface segregation to the concrete implementation in the next chunk, partially because I think structural typing is a road to hell paved with good intentions

PR #744, opened 2025-06-18
feat: impact metrics public api

sighphyre:
Exposes the public API for impact metrics so that it can actually be accessed

I've chosen to nest the impact metrics in a helper property on the client because I think we'll end up with quite a few methods and it's gonna get noisy. If we  hate it later it's easy to collapse.


src/unleash.ts
```diff
@@ -45,15 +45,15 @@ export class Unleash extends EventEmitter {
...
+  protected staticContext: StaticContext;
 
   private synchronized: boolean = false;
 
   private ready: boolean = false;
 
   private started: boolean = false;
 
-  private metricRegistry = new InMemoryMetricRegistry();
+  protected metricRegistry = new InMemoryMetricRegistry();
```
kwasniew, line 56: I'd keep the name explicit about impact metrics. Initially I got confused by the similarity between metrics (current exposure metrics) and metricRegistry. Also I'd stick to plural form "metrics" for the sake of consistency with other names

kwasniew (approved):

PR #747, opened 2025-07-10
feat: impact metrics env resolver

kwasniew:
## About the changes

Impact metrics should report environment from key not static context when possible


sighphyre (approved):
LGTM

PR #758, opened 2025-08-20
feat: fetching mode runtime switch

kwasniew:
## About the changes

This PR introduces the ability to dynamically switch between polling and streaming modes at runtime, allowing the Unleash client to adapt to server capabilities and network conditions without requiring a restart

### Wire Protocol
  1. HTTP Header (Polling → Streaming)
    - Server responds with header: fetch-mode: streaming
    - Client detects this during polling and switches to streaming mode
  2. EventSource Event (Streaming → Polling)
    - Server sends EventSource event: fetch-mode with data: 'polling' (plain string)
    - Client switches back to polling mode

## Discussion points

This PR adds mode switching with minimal code changes however the repository has too many responsibilities. I want to have another PR that splits:
* poling fetch strategy
* streaming fetch strategy
* adaptive fetch strategy changing between the other two

### DIFF
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -217,7 +217,6 @@ export default class Repository extends EventEmitter implements EventEmitter {
   }
 
   async start(): Promise<void> {
-    // Create EventSource if streaming mode and none provided
     if (this.mode.type === 'streaming' && !this.eventSource) {
       this.eventSource = this.createEventSource();
       this.setupEventSource();


src/unleash.ts
```diff
@@ -134,29 +131,6 @@ export class Unleash extends EventEmitter {
         bootstrapProvider,
         bootstrapOverride,
         mode: experimentalMode,
-        eventSource:
```
kwasniew, line 137: moved to the repository

src/repository/index.ts
```diff
@@ -197,6 +217,12 @@ export default class Repository extends EventEmitter implements EventEmitter {
   }
 
   async start(): Promise<void> {
+    // Create EventSource if streaming mode and none provided
+    if (this.mode.type === 'streaming' && !this.eventSource) {
```
kwasniew, line 221: this is a main reason I want to have another PR with separate start strategy for polling and streaming (to remove those if statements)

chriswk (approved):
This is pretty cool. Will it come to our other SDKs as well?

kwasniew:
@chriswk in the first iteration we want to test this option internally and then roll it out to other SDKs if it's useful

src/repository/index.ts
```diff
@@ -476,6 +507,63 @@ Message: ${err.message}`,
...
+    }
+
+    this.emit(UnleashEvents.Mode, { from: 'streaming', to: 'polling' });
+
+    if (this.eventSource) {
+      this.eventSource.close();
+      this.eventSource = undefined;
+    }
+
+    this.mode = { type: 'polling', format: 'full' };
```
sjaanus, line 522: Hard-coding `this.mode = { type: 'polling', format: 'full' }` is bad because it ignores an existing `delta` polling configuration, so carrying over the prior `format` or honoring initial config would prevent unexpected behavior.

sjaanus: Ok, just so we do not forget it, as it is ticking bomb for future.

PR #761, opened 2025-08-22
refactor: extract fetching strategies

kwasniew:
## About the changes

Screenshot 2025-08-26 at 15 23 01

Refactored the repository by extracting fetching strategies into separate, focused classes following the Strategy pattern. 

* PollingFetcher (src/repository/polling-fetcher.ts) - Handles HTTP polling with backoff
* StreamingFetcher (src/repository/streaming-fetcher.ts) - Manages Server-Sent Events streaming
* AdaptiveFetcher(src/repository/adaptive-fetcher.ts) - Coordinates between polling and streaming modes
* FetcherInterface (src/repository/fetcher.ts) - Common interface for all fetchers

Most of the code is move refactoring

### DIFF
--- a/src/repository/adaptive-fetcher.ts
+++ b/src/repository/adaptive-fetcher.ts
@@ -47,9 +47,9 @@ export class AdaptiveFetcher extends EventEmitter implements FetcherInterface {
       return;
     }
 
-    if (newMode === 'polling' && currentMode.type === 'streaming') {
+    if (newMode === 'polling') {
       await this.switchToPolling();
-    } else if (newMode === 'streaming' && currentMode.type === 'polling') {
+    } else if (newMode === 'streaming') {
       await this.switchToStreaming();
     }
   }
@@ -64,7 +64,7 @@ export class AdaptiveFetcher extends EventEmitter implements FetcherInterface {
     this.currentFetcher.stop();
     this.currentFetcher = this.pollingFetcher;
 
-    await this.pollingFetcher.start();
+    await this.currentFetcher.start();
   }
 
   private async switchToStreaming() {
@@ -77,7 +77,7 @@ export class AdaptiveFetcher extends EventEmitter implements FetcherInterface {
     this.currentFetcher.stop();
     this.currentFetcher = this.streamingFetcher;
 
-    await this.streamingFetcher.start();
+    await this.currentFetcher.start();
   }
 
   async start(): Promise<void> {
--- a/src/repository/adaptive-fetcher.ts
+++ b/src/repository/adaptive-fetcher.ts
@@ -6,11 +6,11 @@ import { PollingFetcher } from './polling-fetcher';
 import { StreamingFetcher } from './streaming-fetcher';
 
 export class AdaptiveFetcher extends EventEmitter implements FetcherInterface {
-  private currentStrategy: FetcherInterface;
+  private currentFetcher: FetcherInterface;
 
-  private pollingStrategy: PollingFetcher;
+  private pollingFetcher: PollingFetcher;
 
-  private streamingStrategy: StreamingFetcher;
+  private streamingFetcher: StreamingFetcher;
 
   private options: FetchingOptions;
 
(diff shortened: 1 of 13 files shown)


src/repository/adaptive-fetching-strategy.ts
```diff
@@ -0,0 +1,127 @@
...
+import { UnleashEvents } from '../events';
+import { Mode } from '../unleash-config';
+import { FetchingStrategyInterface, FetchingStrategyOptions } from './fetching-strategy';
+import { PollingStrategy } from './polling-strategy';
+import { StreamingStrategy } from './streaming-strategy';
+
+export class AdaptiveFetchingStrategy extends EventEmitter implements FetchingStrategyInterface {
+  private currentStrategy: FetchingStrategyInterface;
+
+  private pollingStrategy: PollingStrategy;
```
kwasniew, line 11: Both polling and streaming strategies are created upfront rather than on-demand to enable fast mode switching while preserving state (etag, failure counts) across transitions

src/repository/polling-strategy.ts
```diff
@@ -0,0 +1,203 @@
...
+          const fetchingModeHeader = res.headers.get('fetch-mode');
+          if (fetchingModeHeader === 'streaming' && this.options.mode.type === 'polling') {
+            if (this.options.onModeChange) {
+              await this.options.onModeChange('streaming');
+            }
+            return;
+          }
+
+          if (this.options.mode.type === 'polling' && this.options.mode.format === 'delta') {
+            await this.options.onSaveDelta(parseClientFeaturesDelta(data));
```
kwasniew, line 170: this is a change where we can inject onSaveDelta and onSave to each strategy

src/repository/polling-fetcher.ts
```diff
@@ -0,0 +1,203 @@
+import { EventEmitter } from 'events';
+import { parseClientFeaturesDelta } from '../feature';
+import { get } from '../request';
+import getUrl from '../url-utils';
+import { UnleashEvents } from '../events';
+import { FetcherInterface, FetchingOptions } from './fetcher';
+
+export class PollingFetcher extends EventEmitter implements FetcherInterface {
```
kwasniew, line 8: this one has all the polling fetching logic I managed to extract from the repository that was overloaded with responsibilities

src/repository/streaming-fetcher.ts
```diff
@@ -0,0 +1,96 @@
+import { EventEmitter } from 'events';
+import { parseClientFeaturesDelta } from '../feature';
+import { buildHeaders } from '../request';
+import { resolveUrl } from '../url-utils';
+import { UnleashEvents } from '../events';
+import { EventSource } from '../event-source';
+import { FetcherInterface, FetchingOptions } from './fetcher';
+
+export class StreamingFetcher extends EventEmitter implements FetcherInterface {
```
kwasniew, line 9: This is all the streaming logic I extracted from the repository

src/repository/adaptive-fetcher.ts
```diff
@@ -0,0 +1,125 @@
...
+    this.pollingFetcher.stop();
+    this.streamingFetcher.stop();
+  }
+
+  getMode(): Mode {
+    if (this.currentFetcher === this.streamingFetcher) return { type: 'streaming' };
+    return { type: 'polling', format: 'full' };
+  }
+
+  // Compatibility methods for accessing polling strategy internals
```
kwasniew, line 103: repository tests were accessing those so I kept them for now on the adaptive fetcher

src/repository/polling-fetcher.ts
```diff
@@ -0,0 +1,203 @@
...
+  private timer: NodeJS.Timeout | undefined;
+
+  private stopped = false;
+
+  private failures: number = 0;
+
+  private etag: string | undefined;
+
+  private options: FetchingOptions;
```
kwasniew, line 17: we can narrow it down to things we care about

src/repository/fetcher.ts
```diff
@@ -0,0 +1,29 @@
+import { EventEmitter } from 'events';
+import { ClientFeaturesDelta, ClientFeaturesResponse } from '../feature';
+import { Mode } from '../unleash-config';
+
+export interface FetcherInterface extends EventEmitter {
+  start(): Promise<void>;
+  stop(): void;
+}
+
+export interface FetchingOptions {
```
kwasniew, line 10: replace with 2 narrowly focused interfaces for polling and streaming to avoid coupling

src/repository/fetcher.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface CommonFetchingOptions {
+  url: string;
+  appName: string;
+  instanceId: string;
+  headers?: any;
+  connectionId: string;
+  onSave: (response: ClientFeaturesResponse, fromApi: boolean) => Promise<void>;
+  onSaveDelta: (delta: ClientFeaturesDelta) => Promise<void>;
+  onModeChange?: (mode: 'polling' | 'streaming') => Promise<void>;
```
thomasheartman, line 20: I'm still not sure why the individual fetchers need to know about modes. Shouldn't that be entirely left up to the adaptive fetcher? But it looks like your adding some safeguards down the road, so that's probably a good idea. If this is intentional, I'm sure it's good. Just wanted to double check that it actually is.

kwasniew: It is intentional. We can switch modes from the SDK itself by calling setExperimentalMode but also we can switch it from the server driven protocol with either HTTP header for polling or special message for streaming.

PR #766, opened 2025-09-08
Fix switch to polling

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/repository/polling-fetcher.ts
+++ b/src/repository/polling-fetcher.ts
@@ -32,6 +32,7 @@ export class PollingFetcher extends EventEmitter implements FetcherInterface {
   }
 
   async start(): Promise<void> {
+    this.stopped = false;
     await this.fetch();
   }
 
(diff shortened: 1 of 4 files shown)


src/repository/polling-fetcher.ts
```diff
@@ -32,6 +32,7 @@ export class PollingFetcher extends EventEmitter implements FetcherInterface {
   }
 
   async start(): Promise<void> {
+    this.stopped = false;
```
kwasniew, line 35: this was a bug preventing updates after switching from streaming to polling multiple times

src/test/repository.test.ts
```diff
@@ -1938,8 +1938,6 @@ test('setMode can switch from streaming to polling mode', async (t) => {
   t.is(repo.getMode(), 'polling');
   t.true(eventSource.closed);
 
-  await repo.fetch();
```
kwasniew, line 1941: the test will now work without manual fetch trigger and based on the mode change alone

PR #770, opened 2025-10-21
feat: add redis as a backup option

sighphyre:
Very much a draft to see how complex adding Redis as a backup option would be. 100% vibe coded so almost certainly not prod ready until a human has been in the loop

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,11 +435,48 @@ const client = initialize({
 });
 ```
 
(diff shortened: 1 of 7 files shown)


gastonfournier:
We already have an example: https://github.com/Unleash/unleash-node-sdk/blob/75b63da25142a5edd372641299b109e092a62690/examples/custom_storage.js#L7

PR #775, opened 2025-11-20
chore: move metrics api into base Unleash

sighphyre:
We're making this stable. The MetricsClient was there so we could experiment without contaminating our stable public API. Now this needs to get collapsed into the base client

I've chosen to not completely remove the MetricsClient so we can follow semver here. Probably not necessary since we explicitly didn't commit to following semver on that API but it's polite and being polite is nice


src/impact-metrics/metric-client.ts
```diff
@@ -1,137 +1,12 @@
...
-
-    histogram.observe(value, labels);
-  }
-}
+import { Unleash } from '../unleash';
 
 export class UnleashMetricClient extends Unleash {
-  public impactMetrics: MetricsAPI;
-
   constructor(...args: ConstructorParameters<typeof Unleash>) {
```
sighphyre: Just so we can be polite to potential experimental users in the wild

src/impact-metrics/context.ts
```diff
@@ -0,0 +1,17 @@
+import { CustomHeaders } from '../headers';
```
sighphyre: I've given this its own home because I think we need to do a bit more here than we're currently doing. The custom headers function is designed to allow users to hot swap their API key at run time. We're not currently respecting that and we should but I don't want to do that in this PR

src/impact-metrics/metric-client.ts
kwasniew: those users are us :) I appreciate the non breaking migration path so that we can update 2 repos (core + enterprise) at different intervals

PR #782, opened 2025-11-26
feat: bolt in failover

sighphyre:
Bolts in the failover strategy into the streaming fetcher so we can do fail overs. This PR makes an explicit decision to no longer warn on every error, only on errors that cause a failover event - very much intentional, these warnings tend to cause end users to worry that the SDK is broken so we should be as cautious as possible while still being responsible

### DIFF
--- a/src/repository/streaming-fetcher.ts
+++ b/src/repository/streaming-fetcher.ts
@@ -5,16 +5,20 @@ import { resolveUrl } from '../url-utils';
 import { UnleashEvents } from '../events';
 import { EventSource } from '../event-source';
 import { FetcherInterface, StreamingFetchingOptions } from './fetcher';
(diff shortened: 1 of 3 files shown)


src/repository/streaming-fetcher.ts
```diff
@@ -5,16 +5,20 @@ import { resolveUrl } from '../url-utils';
...
 
   private options: StreamingFetchingOptions;
 
+  private readonly failoverStrategy: FailoverStrategy;
+
   constructor(options: StreamingFetchingOptions) {
     super();
     this.options = options;
     this.eventSource = options.eventSource;
+    this.failoverStrategy = new FailoverStrategy(5, 60_000);
```
sighphyre: Config in next PR

PR #786, opened 2025-11-26
chore!: drop node 16 support

gastonfournier:
Node 16 is already end of life, removing this will allow us to unpin some dependencies.

image


gastonfournier:
Close in favor of #789

PR #787, opened 2025-11-26
chore: remove some ts-expect-error and coveralls dependency

gastonfournier:
## About the changes
Try to make our dev life easier by removing unnecessary `// @ts-expect-error` that might be hiding real problems

Also, remove coveralls, which keep the request library pinned:
> `coveralls@3.1.1 requires form-data@~2.3.2 via a transitive dependency on request@2.88.2`

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
     "@typescript-eslint/parser": "^6.0.0",
     "@unleash/client-specification": "^5.1.6",
     "ava": "^5.3.0",
-    "coveralls": "^3.1.1",
     "cross-env": "^7.0.3",
     "del-cli": "^5.0.0",
     "eslint": "^8.44.0",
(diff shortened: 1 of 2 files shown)


src/test/repository.test.ts
```diff
@@ -567,20 +546,20 @@ test('should emit errors on invalid features', (t) =>
...
-      bootstrapProvider: new DefaultBootstrapProvider({}),
+      bootstrapProvider: new DefaultBootstrapProvider({}, 'test-app', 'test-instance'),
       storageProvider: new InMemStorageProvider(),
       mode: { type: 'polling', format: 'full' },
+      refreshInterval: 10,
     });
 
     repo.once('error', (err) => {
       t.truthy(err);
+      repo.stop();
```
gastonfournier: Needed to resolve unhandled promises

PR #796, opened 2025-11-28
Task/migrate off ava to vitest

chriswk:
I decided I'd do the work with some regex and sed statements and some common sense and a nice IDE. Instead of letting AI add custom extensions to the `t` type and making changes I didn't understand. 

Current status is some failing specification tests (in particular 19 and 20) so pinned to 5.1.9. Other than that all tests run green.

### DIFF
--- a/src/test/repository/repository.test.ts
+++ b/src/test/repository/repository.test.ts
@@ -1414,7 +1414,6 @@ test('Stopping repository should stop unchanged event reporting', async () => {
 });
 
 test('Stopping repository should stop storage provider updates', async () => {
-  expect.assertions(1);
   const url = 'http://unleash-test-stop-200.app';
   const feature = {
     name: 'feature',


src/test/global.test.ts
```diff
@@ -39,12 +39,12 @@ test('should be able to call api', (t) => {
   }).on('error', (err) => {
     throw err;
   });
-  t.true(isEnabled('unknown') === false);
+  expect(isEnabled('unknown')).toBe(false);
```
sighphyre: Nice

src/test/index.test.ts
```diff
@@ -26,64 +26,62 @@ const nockRegister = (url: string, code = 200) => nock(url).post(registerUrl).re
...
 
-test('should not return feature-toggle definition if there is no instance', (t) => {
+test('should not return feature-toggle definition if there is no instance', () => {
   // @ts-expect-error
-  t.is(getFeatureToggleDefinition(), undefined);
+  expect(getFeatureToggleDefinition()).toBeUndefined();
 });
 
-test.serial('should start unleash with promise', async (t) => {
+test.sequential('should start unleash with promise', async () => {
```
sighphyre: We still need to do this?

src/test/index.test.ts
```diff
@@ -26,64 +26,62 @@ const nockRegister = (url: string, code = 200) => nock(url).post(registerUrl).re
...
   nockMetrics(url);
   nockRegister(url);
   const unleash = await startUnleash({ appName: 'my-app-name', url });
-  t.truthy(unleash);
+  expect(unleash).toBeTruthy();
   destroy();
 });
 
-test.serial('should start unleash with promise multiple times', async (t) => {
+test.sequential('should start unleash with promise multiple times', async () => {
```
sighphyre: Same as above

PR #800, opened 2025-12-02
chore: update client spec tests to 5.2.2

sighphyre:
Updates the client specification tests to 5.2.2 and patches the SDK to handle the new test cases. We do this by making the fetching layers less opinionated, in that they can now ingest an Unleash compatible response (delta or full) and pass that to the repository layer agnostically. The repository layer is allowed to have opinions on that data because it needs to massage that data into it's own internal shapes

### DIFF
--- a/src/repository/index.ts
+++ b/src/repository/index.ts
@@ -235,28 +235,34 @@ export default class Repository extends EventEmitter implements EventEmitter {
     if ('events' in response) {
       response.events.forEach((event) => {
         switch (event.type) {
-          case 'feature-updated':
+          case 'feature-updated': {
             this.data[event.feature.name] = event.feature;
             break;
-          case 'feature-removed':
+          }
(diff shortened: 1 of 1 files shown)


src/repository/index.ts
```diff
@@ -207,52 +206,74 @@ export default class Repository extends EventEmitter implements EventEmitter {
...
+    } else if ('features' in response) {
+      this.data = this.convertToMap(response.features);
+      this.segments = this.createSegmentLookup(response.segments);
+    } else {
+      this.emit(
+        UnleashEvents.Warn,
+        `Unknown response when applying feature response: ${JSON.stringify(response)}`,
+      );
     }
-    this.connected = true;
```
nunogois: Is this and the `this.stopped` guard no longer needed?

nunogois (approved):
LGTM

PR #807, opened 2025-12-17
Test case reproducing parent child issue

gastonfournier:
## About the test

Creates a parent flag A, which is enabled 50% of the time, with two children at 100%. All are using the same group id and default stickiness.

- When the parent evaluates to false (50% of the time), the child can still be true (because there's a new evaluation)
- When the parent evaluates to true (50% of the time), the child can still be false (because there's a new evaluation)

### DIFF
--- a/src/test/parent-child.test.ts
+++ b/src/test/parent-child.test.ts
@@ -1,0 +1,80 @@
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { mkdirp } from 'mkdirp';
+import nock from 'nock';
+import { assert, expect, test, vi } from 'vitest';
+import { InMemStorageProvider, startUnleash } from '..';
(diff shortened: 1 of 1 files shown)


gastonfournier:
We've decided this is not a bug, each flag evaluation is independent when no session id or user id is provided. Because the parent flag is not 100% rolled out, its evaluation can flip over time (random), so when we evaluate just the parent, we get one value, and when we re-evaluate as part of the child evaluation, we might get a different value, making it look like the evaluation is inconsistent (but it's not).

Why does it work in Frontend API? Frontend artificially injects a session id. And that's actually what we should inject for consistency: the session id works as the consistent seed, removing the randomness, and it aligns the expected behavior. In the same "evaluation session", the child can only be true when the parent is true. The session concept has to be provided externally.

PR #824, opened 2026-02-03
fix: drop negative counter inc

kwasniew:
## About the changes

Spec defines counters as monotonically increasing: https://github.com/Unleash/client-specification/blob/main/impact-metrics/impact-metrics.md#counter


src/impact-metrics/metric-types.ts
```diff
@@ -70,6 +70,7 @@ class CounterImpl implements Counter {
   inc(value?: number, labels?: MetricLabels): void {
     if (isInvalidValue(value)) return;
     const delta = value ?? 1;
+    if (delta < 0) return;
```
kwasniew: after previous line we excluded undefined so it's safest to add negative check here

PR #833, opened 2026-02-16
feat: regex constraint support

sighphyre:
Basic client specification compatible regex constraints

https://github.com/Unleash/client-specification/pull/110

https://linear.app/unleash/issue/2-4186/update-js-sdk-to-handle-regex-cosntraint

### DIFF
--- a/src/test/strategy/strategy.test.ts
+++ b/src/test/strategy/strategy.test.ts
@@ -332,6 +332,23 @@ test('should be disabled when regex is invalid', () => {
   ).toBe(false);
 });
 
+test('should be disabled when regex tries to pass flags', () => {
+  const strategy = new Strategy('test', true);
+  const params = {};
+  const constraints = [
+    { contextName: 'email', operator: 'REGEX', value: '/.*@getunleash\\.ai$/i' },
+  ];
+  const context = {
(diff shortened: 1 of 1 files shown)


src/strategy/strategy.ts
```diff
@@ -90,6 +96,29 @@ const StringOperator = (constraint: Constraint, context: Context) => {
...
+const RegexOperator = (constraint: Constraint, context: Context) => {
+  const field = constraint.contextName;
+  const value = constraint.value as string;
+  const contextValue = resolveContextValue(context, field);
+
+  if (typeof contextValue !== 'string') {
+    return false;
+  }
+
+  const cacheKey = constraint.caseInsensitive ? `${value}i` : value;
```
sighphyre: Hmm, there's a potential collision on the keys here if a case sensitive regex terminates in "i". So example a case sensitive regex of "cati" and a case insensitive regex of "cat". it's probably not going to happen but gosh am I paranoid and I'd rather just make it go away. If we use a prefix tag instead of a suffix tag I think we can eliminate this. Something like this?
```suggestion
const cacheKey = `${constraint.caseInsensitive ? "1" : "0"}:${value}`;
```

PR #858, opened 2026-05-08
fix: correct requests and timeouts on repository, metrics tests

irstavr:
When looking into the migration of 'make-fetch-happen' (https://linear.app/unleash/issue/EG-4361/define-node-sdk-transport-compatibility-contract) and the existing tests, we found some cases where skipped tests would be actually useful in the later phase of migration (or would be nice to have them anyway).

Case-by-case for easier review:

-`repository.test:1026`: wrong comment above test - unskip test, runs successfully.
the 404 test's comment claims `make-fetch-happen` "automatically retries two extra times on 404". Wrong against `node_modules/make-fetch-happen/index.d.ts:62` retries 408, 420, 429, ≥500 and that's it. 404 is not in the list. Un-skip it.

-`repository.test:401,424,447`: 
Tests were skipped because their warn-message assertion is stale. 
They expect responded 502. Waiting for 20ms before trying again. but `polling-fetcher.ts:91-95` emits responded `${statusCode}. Backing off` message - updated it to that instead

-`metrics.test:417`:
The test at src/test/metrics.test.ts:417-456 verifies that when an HTTP 500 is returned, the bucket metrics are preserved so nothing is lost. We were skipping this test even though it is useful. It was failing because of a race condition in the event-emission order inside `sendMetrics()`.
Moved `restoreBucket()` and `metricRegistry?.restore()` to before `configurationError()` / `backoff()`, so the bucket is restored _before_ any warn/error event fires and potentially disables the instance.

-`metrics.test:215`: correct skipped test timeouts 
set correct expected error message

Testfully yours 🧪

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -272,7 +272,12 @@ export default class Metrics extends EventEmitter {
         timeout: this.timeout,
         httpOptions: this.httpOptions,
       });
+
       if (!res.ok) {
+        // restore bucket before any warn/error event fires as it can disable the instance
+        this.restoreBucket(payload.bucket);
+        this.metricRegistry?.restore(impactMetrics);
+
         if (res.status === 403 || res.status === 401) {
           this.configurationError(url, res.status);
         } else if (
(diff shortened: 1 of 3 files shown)


chriswk (approved):
Nice. Great job at bringing back these tests. Goes to show, we need to keep up to date with changes to the libraries we use as well

PR #859, opened 2026-05-11
docs: streaming beta

gastonfournier:
Streaming docs and example


melindafekete:
@gastonfournier thank you! I'll move this to the docs. The readmes we keep short.

PR #860, opened 2026-05-11
task: migrate to pnpm

chriswk:
Since Unleash OSS and unleash-enterprise are now migrated to pnpm, it seems smart to standardise on pnpm in our node repos.


package.json
```diff
@@ -33,13 +33,13 @@
...
     "https-proxy-agent": "^7.0.5",
-    "ip-address": "^10.0.0",
+    "ip-address": "^10.2.0",
     "launchdarkly-eventsource": "2.2.0",
     "lru-cache": "^11.2.5",
     "make-fetch-happen": "^15.0.3",
     "murmurhash3js": "^3.0.1",
     "proxy-from-env": "^1.1.0",
-    "re2js": "^1.2.2",
+    "re2js": "^2.6.1",
```
chriswk, line 42: This one, why did we use such an old re2js version?

sighphyre: It's not old, 1.2.2 was released in Feb. This is a major upgrade, what does the re2js changelog say about the change?

PR #863, opened 2026-05-26
fix: not initialized yet warnings

nunogois:
https://linear.app/unleash/issue/SA-564/fix-not-initialized-yet-warnings-in-the-node-sdk

Fixes "not initialized yet" warnings being emitted in the `getVariant` and `forceGetVariant` methods.

### DIFF
--- a/src/test/unleash.test.ts
+++ b/src/test/unleash.test.ts
@@ -568,9 +568,7 @@ test('should distribute variants according to stickiness', async () => {
   await new Promise<void>((resolve) => {
     unleash.on('synchronized', () => {
       for (let i = 0; i < 10000; i++) {
-        const variant = unleash.getVariant('toggle-with-variants', {
-          someField: genRandomValue(),
-        });
+        const variant = unleash.getVariant('toggle-with-variants', { someField: genRandomValue() });
         // @ts-expect-error
         counts[variant.name]++;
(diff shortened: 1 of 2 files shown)


src/test/unleash.test.ts
```diff
@@ -1077,6 +1077,94 @@ test('should not report dependent feature metrics', async () => {
...
+  instance.getVariant('my-feature');
+  expect(warnings).toHaveLength(1);
+  expect(warnings[0]).toContain("getVariant('my-feature')");
+  expect(warnings[0]).not.toContain('[object Object]');
+  expect(warnings[0]).toContain('"name":"disabled"');
+
+  instance.destroy();
+});
+
+test('should emit correctly serialized warn for getVariant with custom fallback before initialized', () => {
```
sighphyre: I'd yeet this one. I don't think this is an interesting enough problem to cover this deeply with tests. Up to you if you want to keep it

sighphyre (approved):
Cheers

src/test/unleash.test.ts
nunogois: Yep, yeeted: https://github.com/Unleash/unleash-node-sdk/pull/863/commits/8aa433874e21834354292d65545179d3e7603975

PR #864, opened 2026-05-27
chore: fix deprecation notes on MetricsAPI methods via overloads

gergokekesi:
The `@deprecated` JSDoc on `incrementCounter`, `updateGauge`, and `observeHistogram` was attached to the `flagContext` parameter, but with only a single implementation signature TypeScript/IDEs applied the deprecation to the entire method. Every call — including supported ones that omit `flagContext` — was being shown as deprecated.

This change splits each method into two overload signatures:
- A `@deprecated` overload that requires `flagContext` (the one we want to discourage).
- A non-deprecated overload without `flagContext` (the supported usage).

The implementation signature is unchanged. Now only calls that actually pass `flagContext` get the deprecation warning; calls without it resolve to the non-deprecated overload.

The `should increment counter with valid parameters` test now exercises both overloads — one call without `flagContext` (non-deprecated) and one with (deprecated) — and asserts the labels produced by each.

image

### DIFF
--- a/src/impact-metrics/metric-api.ts
+++ b/src/impact-metrics/metric-api.ts
@@ -58,8 +58,11 @@ export class MetricsAPI extends EventEmitter {
   }
 
   /**
-   * @param flagContext - @deprecated This parameter will be removed in a future release.
+   * @deprecated The flagContext parameter will be removed in a future release.
+   *             Call without flagContext.
    */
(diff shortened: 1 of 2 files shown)


FredrikOseberg (approved):
LG

kwasniew (approved):

PR #870, opened 2026-07-14
feat: add SDK flavor and version in register & metrics payload

irstavr:
Right now we can see which SDK and version a client is running, but not what's built on top of it. When someone uses our OpenFeature provider (_or any other integration_) it just looks like a plain Node SDK: we can't tell how much of our usage flows through those integrations, or which versions are running.

This [PR on node-provider](https://github.com/Unleash/unleash-openfeature-node-provider/pull/23) adds the optional `sdkFlavor`, `sdkFlavorVersion` to the Unleash config.
And here we update the SDK config and forwards them in the two payloads we already send: POST /client/register and POST /client/metrics. It goes alongside `sdkVersion` - so we keep the underlying SDK version and gain the integration identity on top.

### What changed
- `UnleashConfig` accepts `sdkFlavor` and `sdkFlavorVersion`
- Unleash passes them through to `Metrics`
- Metrics includes them in `getClientData()` (register) and `createMetricsData() `(metrics). Both are optional and drop out of the JSON body when unset, so nothing changes if you don't set them.

### Testing
- Added Unit tests to cover set and unset scenarios
- Added nock tests to assert the fields actually appear in the register and metrics request bodies and are absent when not configured.

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -395,6 +395,8 @@ export default class Metrics extends EventEmitter {
       platformVersion: this.platformData.version,
       yggdrasilVersion: null,
       specVersion: SUPPORTED_SPEC_VERSION,
+      sdkFlavor: this.sdkFlavor,
+      sdkFlavorVersion: this.sdkFlavorVersion,
     };
 
     if (impactMetrics.length > 0) {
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
Nice!

PR #871, opened 2026-07-15
test: cleaning repo state on `afterEach()` of `repository.test`

irstavr:
Refactors `repository.test.ts's` `afterEach` which had everything commented out (including `nock.cleanAll()`) so that polling/retry not leak across tests (this was found as tests were running for a while in local dev)


sighphyre (approved):
Ye!

PR #872, opened 2026-07-16
chore: update npm-release workflow version to v3.0.2

gastonfournier:
This pins the pnpm version to one compatible with node 20 which we still support


sighphyre (approved):
Thank you!

PR #879, opened 2026-08-06
chore: ignore yarn scripts

kwasniew:
## About the changes

  Dependency install scripts are disabled (ignore-scripts true in .yarnrc). This is the main payload vector in recent attacks. The one postinstall in our tree (esbuild, via vitest) is safe to skip: its binary ships in a platform-specific package, verified working.

  Side effect: yarn 1 can't disable dependency scripts without also disabling our own prepare (husky + build), so:
  - CI jobs get explicit yarn build steps. Previously lib/ was built implicitly by install, and without this test-as-dependency would pack a tarball missing lib/
  - the README tells contributors to run yarn prepare after cloning

  CI now uses --frozen-lockfile. Plain yarn silently accepted lockfile drift. The release template already had this right.

  Renovate waits 7 days before proposing new releases (minimumReleaseAge). Compromised versions are typically removed within hours to days.

  The pre-commit hook calls lint-staged directly instead of via yarn. Husky puts node_modules/.bin on the hook's PATH, so it works regardless of a contributor's package/node manager setup.

  Publishing is unaffected: npm publish reads .npmrc, not .yarnrc, so prepare still builds lib/ at release time.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -6,6 +6,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -16,6 +19,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - name: Setup node
         uses: actions/setup-node@v6
         with:
@@ -33,6 +38,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - name: Setup node
         uses: actions/setup-node@v6
         with:
@@ -55,6 +62,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - name: Setup node
         uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -22,7 +22,9 @@ jobs:
           node-version: ${{matrix.node-version}}
           cache: yarn
       - name: Yarn
-        run: yarn
+        run: yarn install --frozen-lockfile
+      - name: Build
```
kwasniew, line 31: previously it was running automatically from yarn prepare

.github/workflows/build-and-test.yaml
sighphyre, line 87: I think this script needs a patch. It's using npm behind the scenes so it's dodging all the protections in this PR. 

I hate security stuff

sighphyre, line 7: Let's set an explicit read only permission set here too. Node action recommends that anyway: https://github.com/actions/setup-node#recommended-permissions

Think this is probably read only from a higher level permission in the repo or org but being stupidly explicit here is free and nice safety net

sighphyre, line 21: Turns out the credentials are available to other tasks in the job 😢, we can just not do that.

Sorry, can't make a suggestion here because Github seems to be broken (again)

``` js 
with:
    persist-credentials: false
```

sighphyre, line 1: What do the changes in this file do for us? This is built and test, right? So they prevent hostile take over of the CI agent? I think explicit scoping to readonly is the more effective way of doing that.

Changes are nice though, we should do them, just thinking out loud

sighphyre (approved):
Awesome

PR #881, opened 2026-08-14
chore(deps): bump re2js to ^2.3.2

sjaanus:
## What

Bump `re2js` from `^1.2.2` to `^2.3.2`.

## Why

The SDK is the only thing still on `re2js` 1.x. `unleash-server` already depends on `re2js@2.3.2`, so any consumer that pulls in **both** the SDK and the server resolves two major versions of `re2js` and loads **two full copies of the RE2 engine** into memory. A heap snapshot of a live Unleash Cloud pod showed the re2js source present twice (~0.6–1.3 MB of duplicated string + compiled code per process).

Aligning the SDK on 2.x lets those consumers dedupe to a single copy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "make-fetch-happen": "^15.0.3",
     "murmurhash3js": "^3.0.1",
     "proxy-from-env": "^1.1.0",
-    "re2js": "^1.2.2",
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Cheers!

